PR #7, opened 2015-05-14
add labelField prop option

beeant:
add labelField option for the property name of `tags collection` to render as an option / item label (default to `text`

### DIFF
--- a/lib/Tag.js
+++ b/lib/Tag.js
@@ -24,9 +24,15 @@ var styles = {
 var Tag = React.createClass({
     mixins: [ReactDND.DragDropMixin],
     propTypes: {
+        labelField: React.PropTypes.string,
         onDelete: React.PropTypes.func.isRequired,
         tag: React.PropTypes.object.isRequired,
         moveTag: React.PropTypes.func.isRequired
+    },
+    getDefaultProps: function() {
+        return {
+            labelField: 'text'
+        };
     },
     statics: {
         configureDragDrop: function(register) {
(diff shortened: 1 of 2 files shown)


prakhar1989:
Hi @beeant! Can you please tell me how this is useful? I'm probably missing something and hence not seeing an immediate use of this? Thanks for the PR, though!

beeant:
thank you for your response.
this is useful if you have set of collection data and you want to show the string inside the key property set in labelField. For example:

```javascript
var data = [{
  lower: "a",
  upper: "A"
}, {
  lower: "b",
  upper: "B"
}]:
```

with labelField option, you can set `lower` or `upper` to be shown.

without labelField, you can only show `text`

prakhar1989:
Makes sense, @beeant. I had plans of giving more flexibility to the `suggestions` data, and this seems like a good fit. I'm not sure if I like the `labelField` term, but I'll merge this PR and change it later. I hope that's okay. I hope the component is useful to you :) Thanks!

beeant:
@prakhar1989 thanks for the merge.
I take the labelField naming from selectize (http://brianreavis.github.io/selectize.js/)
I've been using selectize with angular.js, it seems to be matured, many ways to do customisation. but just it has jQuery dependencies.

prakhar1989:
Ah ok. Didn't know about selectize. Looks good. Thanks for the link. :+1:

prakhar1989:
> Use your left/right arrow keys to move the caret (ibeam) between items. (from selectize)

I really like this functionality. Looks much more usable than dragging items. What do you think?

etosch:
@prakhar1989 : was this last comment meant for me or @beeant?

prakhar1989:
Ah sorry for the confusion - it was meant for @beeant :)

etosch:
Gotcha! I was following the thread and it occurred to me that you might be talking about something surveyman-specific, rather than a general design. (FWIW I always prefer using the keyboard over using the mouse, but I am probably in the minority of end-users in that way).

prakhar1989:
> (FWIW I always prefer using the keyboard over using the mouse, but I am probably in the minority of end-users in that way).

Agree with you completely. That's the main reason why I really like the way `selectize` handles it. But since, you mentioned that the key users of surveyman are going to non-programmers, I felt that dragging might be more intuitive. If time permits, I'll really like to add this functionality as well.

Issue #16, opened 2015-09-17
Prevent Delete on Text Input Delete


jkusachi:
With some updates to the CSS, one is able to do a 'stacked' approach of the tags rather than have the tags inline with the input field.

![screenshot 2015-09-17 10 19 08](https://cloud.githubusercontent.com/assets/332474/9940596/a5c654e4-5d25-11e5-8565-6f035143fc22.png)

In this case however, and tags exist, pressing the "Delete" key on an empty Input Box will delete the most recently added Tag.  This is not always desired.

Suggestion: add a property `allowDeleteFromInput` that has either `true` or `false` that controls this.  

I can create a PR if you'd like

prakhar1989:
Makes sense. However, rather than creating an additional parameter did you try the following - 

```javascript
handleDelete: function(i) {
    return;
},
```
So in your `handleDelete` function, just don't do anything. I believe this should solve the problem.
Let me know if you face issues.

jkusachi:
thanks @prakhar1989 

in this scenario though, users wouldn't be able to click the 'X' on the button to delete a tag

in the case presented, I would only want to disable it for the onKeyDown on empty input field.

dauidus:
+1

I'm looking to do this, too.  I've styled the tags to mimic the default tag input field in WordPress (I'm using this in a plugin).  In @jkusachi's scenario, this would allow me to mimic the functionality of the default input field, too.

-D

PR #17, opened 2015-09-21
Feature/allow delete from input prop

jkusachi:
PR for Issue: #16 

Adding new optional prop `allowDeleteFromEmptyInput` (default: `true`) to toggle whether or not tags should be deleted when the 'Delete' key is pressed on an empty Input

This is useful when tags are not inline, but 'stacked'

screenshot 2015-09-17 10 19 08


prakhar1989:
Thanks for the PR! Merged: )

Issue #18, opened 2015-09-25
Allow this.props.onChange for API calls


checkhavoc:
Any chance this can be added to the input onChange handler? This would be nice to have so that the input can make API requests just in case suggestions are not already pre-loaded. If there is a better way to handle this type of request, please let me know. Thanks!

```javascript
handleChange: function handleChange(e) {

        // Check if this.props.onChange exists
        if (this.props.onChange){
            this.props.onChange(e.target.value.trim())
        }
        var query = e.target.value.trim();
        var suggestions = this.props.suggestions.filter(function (item) {
            return item.toLowerCase().search(query.toLowerCase()) === 0;
        });

        this.setState({
            query: query,
            suggestions: suggestions
        });
    },
```

Jayr Camacho

prakhar1989:
Fixed in #19

atley:
I dont get it. I have a Search API for tags. I want to update the suggestion Array onChange. How do I do that?

PR #19, opened 2015-09-29
Add handleInputChange event handler

checkhavoc:
PR for issue #18 

Added an optional prop (handleInputChange) for more flexibility regarding suggestions being passed in. This will allow for custom API requests to be made on the input's onChange, which will make suggestions more dynamic


prakhar1989:
Looks great! Thanks for the PR! :beers:

PR #20, opened 2015-09-29
use node_modules/.bin/babel

theverything:
Using the node_modules .bin is more future proof. If the babel project
were to rename or change their file structure the build script would
fail.

As a matter of fact they did this already.
`node_modules/babel/bin/index.js` has been renamed to `babel.js`


prakhar1989:
Ah! I had no clue about this. Thanks!

PR #22, opened 2015-10-27
React 0.14 + react-dnd 2.0

izziaraffaele:


prakhar1989:
This is awesome! Thanks a ton @izziaraffaele. I'd been wanting to bump up dep versions but couldn't find the time. :beers: 

I see you've removed the babel transform and instead added ` <script type="text/babel" src="main.js"></script>`. How does `text/babel` work? I probably might have missed news on that one.

izziaraffaele:
@prakhar1989 the JSX transformer was deprecated in favor of babel. Babel has an in-browser transformer that I included in the html as well that acts more or less as the old JSX transformer.

For more info https://github.com/facebook/react/blob/master/CHANGELOG.md#react-tools--babel

prakhar1989:
Great! Thanks for the clarification! I'll test this branch over the weekend, and release a new version on NPM. Thanks for putting this together again. Much appreciated :+1:

izziaraffaele:
awesome! :beers:

PR #26, opened 2015-11-23
Some updates.

xonev:
* Configure min query length.
* Update suggestions when suggestions prop is updated.
* Allow tags to show up not inline with the input element.
* Remove unnecessary `findDOMNode` calls.


prakhar1989:
This looks great! Thanks a ton for this PR!

PR #27, opened 2015-11-25
Allow to specify any keys as tag delimeters

not-raspberry:
It's possible now to specify keys that, when pressed, apply current input. Previously those keys were hardcoded. Default behaviour does not change - enter and tab still terminate the input by default.

Now exporting ``Keys`` to allow library users to reuse some of react-tag-input codes.

Also added ``minQueryLength`` to README because it was missing (separate commit).


not-raspberry:
I commited with the wrong email. Fixing now.

not-raspberry:
OK, we're good.

prakhar1989:
Looks great! Thanks @not-raspberry

PR #29, opened 2015-11-26
bugfix: user input was interpreted as a regular expression

not-raspberry:
It would crash if the input was an invalid regular expression, like 'aaa['. It would match everything if the input was '..' or '.*'. Now it's interpreting the input as a string.


prakhar1989:
Nice catch! Thanks!

PR #30, opened 2015-11-27
build it so people using Github dependency can benefit too

not-raspberry:
I forgot to build it yesterday.

Pushing built code to Github is not the best practice but I suppose there's a number of people using the Github URL as their dependency (they probably haven't caught up with the fact you just pushed a new version to npm).


prakhar1989:
Hi @not-raspberry 

> people using the Github URL as their dependency

I don't support this since this is against Github's terms of service. FWIW, I have pushed the latest changes on NPM so this should be fine now. Appreciate the PR, but I'm afraid I can't merge it.

PR #34, opened 2016-01-11
Do not use regular expressions with user input.

xonev:
There was still a place that was calling `search` on the user's input. This should be completely fixed now.


prakhar1989:
Thanks :+1:

Issue #32, opened 2015-12-14
Allow to pass style classes as input


eoinmurray:
Consider something like 

```
<ReactTags
   ...
   classes = {{
      input: 'some-input-class',
      tagInput: 'some-other-class'
   }}
/>
```

I think it would make it much easier to integrated into existing projects that have existing UI packages in place

bitdivision:
I would also appreciate this. Using hard-coded class names is not really ideal in many situations.

I may be extending this module over the coming days. Is it worth making a PR for this?

prakhar1989:
@bitdivision A PR definitely sounds great! The ability to configure class-names with reasonable defaults would be useful. Let me know if you need any help.

daamsie:
I tried this with tagInput: 'form-control' for bootstrap styling. However the tagInput class is not on the actual input itself, but on a div around the input. That to me is not expected behaviour. 

But more importantly it looks like the there isn't currently a way of setting the class of the input itself at the moment.

PR #35, opened 2016-01-14
Allow user to specify custom class names on the component

bitdivision:
Fixes issue #32.

One potential issue is that this will not merge the defaultProps with the passed class names. I'm not sure if this is a probable use-case but can add if required.


prakhar1989:
This looks great!

> One potential issue is that this will not merge the defaultProps with the passed class names.

Let me know if I getting this right - you mean the user will have to either pass all custom classes or stick with the default ones right? I think it makes sense if we merge the default class object with the custom one so that the user can use the lib with minimum configuration. 

PS: [Lodash](https://lodash.com/docs#merge) already comes with a merge function that you can use.

bitdivision:
Yeah, that's exactly what I meant. I hadn't noticed that lodash was already in there. I'll fix it up this evening probably.

prakhar1989:
Sure :+1:

cklab:
@bitdivision any plans to wrap up the PR?

bitdivision:
@cklab 

Sorry for the delay, will finish this off tomorrow.

sapegin:
@bitdivision @prakhar1989 It would be ver cool to have this merged.

bitdivision:
@prakhar1989 I've added the functionality to merge default classnames. It's not the cleanest solution, since it stores classnames in state, but I couldn't see a nicer way of doing it.

Sorry about the rebase, probably should have merged instead.

prakhar1989:
@bitdivision Don't worry about it. Thanks for the PR :wink:

PR #37, opened 2016-01-21
Don't use String.prototype.startsWith (yet)

rnicholus:
I noticed you are using this in reactTags, specifically when the determination is made as to which suggestions to display in the menu. Unfortunately, `String.prototype.startsWith` is not yet well supported. By using this, you're eliminating support for anything older than Microsoft Edge and Safari 9. Also PhantomJS will choke as well, since that is based on an old version of WebKit.

Simply using `indexOf` and checking for a value of `0` will give you the same result as `startsWith`.


prakhar1989:
Ah! I wasn't aware of this. `startswith` is a rather recent change due to a PR. Thanks for pointing this out!

rnicholus:
Thanks for merging this. Are you going to `npm publish` soon, or perhaps you already have?

We are evaluating your library for use in our enterprise app. If we end up using it, I'll probably open up a pull request for some unit tests, the one big thing that is missing from your library.

prakhar1989:
I was waiting for a few more fixes to land before I publish on NPM. 

> I'll probably open up a pull request for some unit tests
It would be fantastic if you add some tests. It has been on my list of things to do but unfortunately grad school has been taking all by time. 

Thanks!

rnicholus:
It's easy to publish to npm, and incrementing the hotfix version is no big deal, IMHO. The adjustment I've made prevents your library from breaking in all versions of IE, anything other than the latest Safari, and stock Android. With this in mind, you may want to just publish and save the other fixes for 2.0.6. Just my 2 cents. Thanks for the awesome library.

prakhar1989:
Makes sense. I just published the version on npm :+1:

rnicholus:
Thanks so much!

PR #38, opened 2016-01-22
Allow custom remove/delete component to be specified

rnicholus:
We needed to use an icon for the remove button instead of the supplied "x". Also, since an anchor element with no href is (usually) invalid, and an anchor is not really appropriate for actions that do not update the URL, we wanted to use a `<button>` element for the delete/remove element instead. The best way to accomplish this is to ensure the library allows the entire remove/delete element to be replaced. 

So, I added an optional property to `<ReactTags />` - `removeComponent`. It expects a React component (which is a `function`), and passes it to the `<Tag />` component. This property is optional. If omitted, the existing anchor with "x" text node is used. Otherwise, the passed component is rendered in its place. Also, all attributes/properties that would normally be added to the anchor element are passed on through to the custom remove component if it is supplied.

If you want the remove/delete element to render as it always has, no code has to change for the user/integrator. But if you want to supply your own, your code would look something like this:

```javascript
import {WithContext as ReactTags} from 'react-tag-input'

class Foo extends React.Component {
   render() {
      return <ReactTags removeComponent={RemoveComponent}/>
   }
}

class RemoveComponent extends React.Component {
   render() {
      return (
         <button {...this.props}>
            <img src="my-icon.png" />
         </button>
      )
   }
}
```

The "ReactTags__remove" and `onClick` handler properties will be included on the `<button>` by using the [JSX spread attribute](https://facebook.github.io/react/docs/jsx-spread.html).


prakhar1989:
This looks great, Ray! Thanks. One quick thing - can you send in a commit with appropriate changes to the README too so that this useful feature is documented? Following this, I'll merge and `npm publish` right away.

rnicholus:
Sure thing. I'll type up something now.

rnicholus:
README updated in 0473cdcc40c7e22a533337cfcb31250bdcf490df

prakhar1989:
Brilliant! Merging :beers:

Issue #39, opened 2016-01-26
Automatically choose option on trigger key (enter/tab/etc) if only one suggestion matches


rnicholus:
During integration of react-tags into our app at Widen, we had a requirement that a proposed tag should be accepted as the matching suggestion if only one match can be found. This would happen when the user hits enter or tab or whatever the trigger key is set to.  For example:

1. Suggestions of "one, two, three"
2. User starts to type "th"
3. User hits enter.
4. Tag is persisted as "three".

This mirrors behavior of other in-the-wild tagging systems, such as what you might see on Slack.

I realize that this should probably not be the default behavior for a library like this, but perhaps as an option? Just asking if you are interested in a PR will this code. If so, I'm interested in what thoughts you might have regarding how this feature should be enabled. 

If you are not interested, no big deal. We have implemented this in our wrapper of react-tags.

jde:
+1 This may seem like a small ui improvement, but when someone is doing a lot of work on a system things like this have a huge productivity boost.  Also, as op mentioned, it is in line with expected behavior so no user training necessary!

prakhar1989:
Sorry for the late reply. This surely sounds more user-friendly than what's currently in place. If you've already have this implemented, send in a PR and I'll be glad to merge it in.

rnicholus:
My plan is to introduce a new option - `autocomplete`. Thee possible values:

- `false` (default) - component functions as it did now
- `true` - when trigger key (such as enter) is pressed, first matching suggestion is used.
- `1` - when trigger key (such as enter) is pressed, matching suggestion is used only if there is a single matching suggestion

This new option/behavior has no effect if there are no `suggestions`. I'll put something together shortly. Note that I also plan on including an a11y fix we came across during accessibility testing.

rnicholus:
Implemented in #40 and part of 2.2.0.

patrick-samy:
Is it possible to avoid adding the tag if there is no suggestion? Effectively limiting the tags values to the suggestion list.

PR #40, opened 2016-01-28
Autocomplete

rnicholus:
Two things in this pull request:

1. Addresses #39 
2. An a11y fix for the text input. A `placeholder` is not enough to make it accessible, so now an `aria-label` attribute mirrors the `placeholder`.


prakhar1989:
You're awesome, Ray! Merging :dancer:

Issue #33, opened 2016-01-01
Warning: Each child in an array or iterator should have a unique "key" prop


beeant:
I encountered this problem when I'm trying to add new fresh tags (for example the tags is not saved in database, they don't have tag id)

So I'm thinking if we can add:
```javascript
{tag.id || i}
```
in here
https://github.com/prakhar1989/react-tags/blob/master/lib/reactTags.js#L188

I can add `tag.id` to all the new fresh tags, but I have to clean it back when sending them to the server. Because if I don't it will have conflict with the existing tags with same id.

Thank You

prakhar1989:
Makes sense. Would you be interested in sending a pull request with this change in place? I'll be happy to merge that in!

prakhar1989:
Oh and Happy New Year! :fireworks:

beeant:
thank you for the merge! :)

cklab:
Running into the same warning every now and again. Scenario is:

Enter tag with `{ name: "tag", id: 1 }` and then a new, non-existent tag `{ name: "New Tag" }`

"New Tag" will get the same "key" field.

Generating ID should be consistent and deterministic, IMO ignore user input when generating a key. Submitted PR #54

francisrod01:
The same error.but my default tags are `['tag 01', 'tag 02', 'tag 03']` without a key.

PR #50, opened 2016-03-07
Fix warning: Each child in an array or iterator should have a unique key

beeant:
Fixes issue #33

Set tag entry id as key for newly added tags (for example the tags is not saved in database, they don't have tag id)


prakhar1989:
Thanks!

PR #54, opened 2016-03-14
use index for key generation for <Tag>s #33 

cklab:


prakhar1989:
Hi, can you explain why `index` is better than `tag.id`?

cklab:
@prakhar1989 Sure, the `index` is guaranteed to never collide whereas `tag.id` potentially can (especially so during the creation of new tags that are submitted to an external API in batches instead of one-by-one).

I tried to outline the use case & problem I ran into under issue #33 (https://github.com/prakhar1989/react-tags/issues/33#issuecomment-196472809)

PR #61, opened 2016-04-27
Added new prop: shouldRenderSuggestions.

MartynasZilinskas:
Added prop `shouldRenderSuggestions`. By default it works with `minQueryLength` prop. It's more convenient to tell when exactly show suggestions. With this prop it creates more flexibility. 

Note: All Javascript files ".js" that contains JSX should be named ".jsx".


prakhar1989:
Looks great! Thank you for your contribution 🌟

prakhar1989:
@MartynasZilinskas : I just tried the running the build and it seems it no longer works. I believe your `.jsx`  and plugin addition requires more changes in `package.json`. Any idea what's wrong?

PR #75, opened 2016-06-05
Use ✕ instead of x for default remove component

tommoor:
Changes the default remove component to use a proper cross (✕) symbol instead of the letter "x".


prakhar1989:
Thanks for your contribution @tommoor 👍

PR #76, opened 2016-06-05
stopPropagation of ESC key

tommoor:
If `react-tags` catches the ESC key to close the suggestions menu it should not be allowed to propagate into the parent application.


prakhar1989:
Nice catch!

PR #78, opened 2016-06-13
Updates onPaste PR

bmordan:
Merges changes from the HEAD of prakhar1989/react-tags


prakhar1989:
Looks great! Just tried it out too :) 

Thanks for your contribution, @bmordan!

PR #87, opened 2016-07-06
Clearing input once invoked onBlur handler

Wildhoney:
Sorry &mdash; a little oversight!

I think once we've invoked the `handleInputBlur` function, the `input` should be cleared, as the assumption is that the text was added as a tag.

Alternatively we could clear the input based on the return value of the `handleInputBlur` function, but that seems like it could cause confusion, although it would offer the developer greater control.

### DIFF
--- a/lib/ReactTags.js
+++ b/lib/ReactTags.js
@@ -111,7 +111,8 @@ var ReactTags = React.createClass({
   handleBlur: function(e) {
     var value = e.target.value.trim();
     if (this.props.handleInputBlur && value.length){
-      this.props.handleInputBlur(value)
+      this.props.handleInputBlur(value);
+      this.refs.input.value = "";
     }
   },
   handleKeyDown: function(e) {
--- a/lib/Suggestions.js
+++ b/lib/Suggestions.js
@@ -30,7 +30,7 @@ class Suggestions extends Component {
     const suggestions = props.suggestions.map(function (item, i) {
       return (
         <li key={i}
-        onClick={props.handleClick.bind(null, i) }
+        onMouseDown={props.handleClick.bind(null, i) }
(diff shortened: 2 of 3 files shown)


Wildhoney:
I've also made a second change, which is changing the `onClick` on the list of suggestions to an `onMouseDown` &ndash; this allows `onclick` suggestions to first perform its action before `onblur` of the input. By doing this it allows clicking on a suggestions to take precedence. As the `handleSuggestions` function sets the `input` to `null`, the `onblur` event won't perform *its* action.

_Without the changing the `onblur` was taking precedence over the `onclick` suggestions, which is far from ideal, as the user click is a clear intent._

Changing the `onClick` to a `onMouseDown` should be perfectly backwards compatible, as far as I understand 👍

Wildhoney:
Any update on this PR?

prakhar1989:
Whoops! Totally forgot about this - really sorry! I'll get back to you on this by evening :)

prakhar1989:
Looks good to me! Thanks for catching this and sending in the fix! :)

Wildhoney:
Thanks!

Wildhoney:
Published? 🙈

prakhar1989:
Just published. Sorry, I forgot again.  😿

cjcheshire:
So I suspect changing from onclick is what broke mobile as the suggestion list isn't working on the iPhone and possibly Android too. It remain open not accepting interaction. :/

Wildhoney:
Possible to use `ontouchstart` and `ontouchend`?

cjcheshire:
Interestingly reading more around it we may be able to use ```cursor: pointer``` within the css of the li.  A quick hack test on my iPhone simulator proved successful.  

We theme the selector but it might be useful to add to the default css. :)

Issue #73, opened 2016-05-31
make filteredSuggestions settable by user


facultymatt:
It I have a list of suggestions like so: 

cat
cats
cars
cards

And i start to type

`a`

then I will get a suggestion box with all 4 suggestions and the `a` characters underlined (marked). 

If I then type `ar` 0 suggestions will render. 

It I type `car` then 2 suggestions will render.

I would expect that  any suggestion matching the input pattern would render, not just if the suggestion matches on the first characters.

facultymatt:
I discovered this is because of https://github.com/prakhar1989/react-tags/blob/master/lib/reactTags.js#L81-L85 and maybe other functions which call this.

This behavior makes sense sometimes, but in all cases its not ideal. For example in my case the filtering is done via an API call so my suggestions list is ready to go - the extra filtering step produces unexpected results. 

Is is possible to make this an option so the user can provide their own filter function? What are some other good solutions?

facultymatt:
Noticed that it was easy to override `filteredSuggestions` - would be great to add to the documentation as a supported option 😄

prakhar1989:
Sounds like a good idea. I feel that making the `filter` function, customizable would be useful, but I'm not sure how the API should look like. My understanding is that most users would **not** want to customize that function and hence there should be a reasonable way of making the default available.

I'll see how others are doing it and in the meantime if you have any ideas, feel free to share.

PR #91, opened 2016-07-28
Add props.filterSuggestion

malantonio:
This was being discussed in #73 and as I was writing up a comment on how it could be implemented, it seemed easier to just provide a PR.

This adds a `filterSuggestion` prop for a function that's passed to `suggestions.filter`. The query value is bound, so the signature becomes:

```javascript
function filterSuggestion(query, suggestion, index, suggestionArray) { /*...*/ }
```

I've also set the `defaultProp` to match the original filter function, so the original behavior remains.

Let me know what you think; I'm happy to make changes where needed!


prakhar1989:
Hi @malantonio, thanks for sending in the PR. I think this looks great, but there's parallel work going on in #92. I would recommend that you give that a look and see if that sounds like a good implementation. Also, feel free to help out in that PR, if you so prefer. Thanks :)

PR #93, opened 2016-08-05
Auto-scroll to suggestion on up/down arrow

rnicholus:
- [x] Auto-scroll on up/down arrow (only if needed).
- [x] Update package.json to v4.2.0.

If the suggestions container is given a `max-height` & an `overflow-y: auto`, then the active suggestion outside of the viewable area will be hidden if selected via up/down arrow. While the suggestions list has an unbounded height by default, it seems likely that many would want to limit this in some way, as a very long suggestions list may look unpleasant in some scenarios, such as when the tags component is rendered inside of an element with a restricted height.

This change is not opt-in, but it should only have an effect if the list is height-limited in some way. So it seems appropriate to make this a default behavior. This way, the component functions as expected without having to hunt for a configuration option to set. But I understand if you want this to be opt-in anyway. Let me know @prakhar1989. This change can target v4.2.0.


lib/Suggestions.js
```diff
@@ -12,6 +25,14 @@ class Suggestions extends Component {
     classNames: React.PropTypes.object
   }
 
+  componentDidUpdate = prevProps => {
+    const suggestionsContainer = this.refs.suggestionsContainer
+    if (suggestionsContainer && prevProps.selectedIndex !== this.props.selectedIndex) {
+      const activeSuggestion = suggestionsContainer.querySelector('.active')
+      activeSuggestion && maybeScrollSuggestionIntoView(activeSuggestion, suggestionsContainer)
```
stephenburgess8: While this is very pithy, I think it would be clearer to just say 
```
if(activeSuggestion) {
    maybeScrollSuggestionIntoView(activeSuggestion, suggestionsContainer)
}
```

lib/Suggestions.js
```diff
@@ -29,12 +50,12 @@ class Suggestions extends Component {
     const { props } = this;
     const suggestions = props.suggestions.map(function (item, i) {
       return (
-        <li key={i}
-        onMouseDown={props.handleClick.bind(null, i) }
-        onMouseOver={props.handleHover.bind(null, i) }
-        className={i == props.selectedIndex ? "active" : ""}>
-        <span dangerouslySetInnerHTML={this.markIt(item, props.query) } />
-        </li>
+          <li key={i}
```
stephenburgess8: Would we expect the options of this dropdown to change dynamically? It seems like a reasonable assumption. Can use `item` as the key, or does `item` have any properties that can be used as a key?

https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318#.woflu8at1

lib/Suggestions.js
```diff
@@ -29,12 +50,12 @@ class Suggestions extends Component {
...
+              onMouseDown={props.handleClick.bind(null, i) }
+              onMouseOver={props.handleHover.bind(null, i) }
+              className={i == props.selectedIndex ? "active" : ""}>
```
stephenburgess8: If we were to use a string instead of an index for the string, this line could stand to be spread out over multiple lines for clarity. This is where the magic is happening anyway so good to make that very readable.

lib/Suggestions.js
rnicholus: I think changes to this code might be out of scope. Looks like my ide updated the indentation here, but that wasn't intentional.

lib/Suggestions.js
rnicholus: Changes here should be considered in a different case, in my opinion, as this pull request focuses on the scrolling issue specifically.

lib/Suggestions.js
rnicholus: I can adjust for clarity.

prakhar1989:
Heya guys, thanks for taking on this! I agree with your suggestion that should be the default functionality. Also, I was wondering if it would be possible to add tests for this feature.

The rest of it looks good to me.

rnicholus:
I thought about that, but it seemed like the extreme amount of work would bring little benefit. In order to properly unit test this change, we would have to limit the height of the suggestions box with CSS (as part of the test), provide a set of suggestions, enter a few characters that result in a number of suggestions that exceeds the height of the suggestions box we set earlier, then programmatically use the arrow keys to navigate the suggestions, moving to the suggestion off the screen, and then somehow determine if the suggestion is properly visible. I'm not quite sure how this can be verified programmatically in a headless environment. It's probably possible, but likely a lot of work. Manual verification seems most appropriate here. 

I verified locally w/ the example app (which I temporarily modified using the conditions outlined above), and can run further verification on our internal project as well.

prakhar1989:
That sounds good! Is there anything else pending on this PR? If not, you can go ahead squash and merge, after which I'll publish this change.

rnicholus:
Great! After I run some final tests on our internal project, I will do squash+merge.

rnicholus:
Initial testing on Chrome, FF, Safari, IE11, and Edge look good - merging.

rnicholus:
Once you get a chance to `npm publish` 4.2.0, I'll be sure to delete this branch once we've updated our internal code.

prakhar1989:
Done!

khoramshahy:
hi 
this feature dose not work for me
should I add any prop to component to get it work?
this is my code:

`
<ReactTags
                                                id={`${key} + '-items-include'`}
                                                autofocus={activeTagInput.indexOf('items-include') > 0 ? true : false}
                                                inline
                                                inputFieldPosition="top"
                                                tags={media[key].items.include}
                                                suggestions={media[key].items.suggestions}
                                                handleDelete={handleDelete}
                                                handleAddition={handleAddition}
                                                handleInputFocus={(value) => handleInputFocus(key + '-items-include', value)}
                                                handleInputChange={(value) => handleInputChange(key + '-items-include', value)}
                                                autocomplete={true}
                                                labelField="uri"
                                                minQueryLength={0}
                                                inputValue={activeTagInput.indexOf('items-include') > 0 ? media[key].items.searchValue : ""}
                                                delimiters={delimiters}
                                            />
`

PR #95, opened 2016-08-10
Fixed bug with pressing keys Up and Down when suggestion list is empt…

landrew31:
…y for current input value.
There was a problem - when we input some tags, and then press Up or Down buttons (accidentally or deliberately) and then once press Enter, our tag isn't saved into array in state.


Wildhoney:
Could you create a test for this scenario, please?

prakhar1989:
@landrew31: Since there hasn't been any update on this, I'm closing this for now. Feel free to re-open as and when you're ready.

PR #96, opened 2016-08-12
Fix/select remaining suggestion

whitmanschorn:
This fixes the following issue:

> Type a few letters, highlight a suggestion on the list second down or farther, and keep typing so that the suggestion disappears. `selectedIndex` will be out of bounds of the suggestions array. (And hitting enter will add an `undefined` tag 💀)

This ensures that `selectedIndex` is adjusted to the last available suggestion while typing. Also added a test. 

Also updated package.json to v4.2.1.

P.S. this library is great, I'm happy to have a chance to contribute!


prakhar1989:
Heya @whitmanschorn, this is awesome and it looks great! Thanks for adding the tests and fixing this bug. I'll merge this in and release this on NPM right away.

Thanks a ton!

PR #99, opened 2016-08-17
Asserting that the shift key isn't being held down

Wildhoney:
In [the example](http://prakhar.me/react-tags/example/) holding down <kbd>shift</kbd> whilst pressing <kdb>enter</kbd> adds the tag, when in reality that key combination is technically different from just <kbd>enter</kbd>.

For the above example it doesn't matter too much, but when you set your `delimiter` to <kbd>tab</kbd> for instance, then you're unable to enter <kbd><</kbd> into the textbox because the `keyCode` for <kbd>,</kbd> is the same as <kbd><</kbd> and thus takes it as the `delimiter`.

This change ensures the <kdb>shift</kbd> isn't being held down. Interestingly I couldn't find any tests for `addTag` and so I didn't add a test for this PR. Existing tests pass just fine.

I don't imagine this change will affect anybody as it really is an edge-case, however it is a breaking change.

### DIFF
--- a/lib/ReactTags.js
+++ b/lib/ReactTags.js
@@ -141,7 +141,7 @@ var ReactTags = React.createClass({
     // When one of the terminating keys is pressed, add current query to the tags.
     // If no text is typed in so far, ignore the action - so we don't end up with a terminating
     // character typed in.
-    if (this.props.delimiters.indexOf(e.keyCode) !== -1) {
+    if (this.props.delimiters.indexOf(e.keyCode) !== -1 && !e.shiftKey) {
       if (e.keyCode !== Keys.TAB || query !== "") {
         e.preventDefault();
       }


prakhar1989:
Sorry for the late update on this, Adam. I was travelling so didn't get a chance earlier to look at this. 

> but when you set your delimiter to tab for instance, then you're unable to enter < into the textbox because the keyCode for , is the same as < and thus takes it as the delimiter

Pardon my ignorance here, but how does adding the check you added (i.e. removing shift-key) fix this problem? 

Thanks for the PR, though :)

Wildhoney:
Sorry, it should read "when you set your delimiter to <kbd>,</kbd> for instance".

Comma and <kbd><<kbd> share the same `keyCode` &mdash; alternatively use [http://caniuse.com/#search=KeyboardEvent.key](`KeyboardEvent.key`).

prakhar1989:
Ah ok! Thanks for clarifying! :)

Wildhoney:
Thanks!

PR #100, opened 2016-08-20
add inline styles for background and text color - see README

stahlmanDesign:
I needed to change the tag colour dynamically so I couldn't use classes (don't know the colour ahead of time). To avoid using JQuery or another direct DOM manipulation technique, I added optional properties to ReactTags to set the backgroundColor and text colour. I did a build and tested it. It works if you set the properties, and it reverts to defaults if you do not set them.


prakhar1989:
Hey @stahlmanDesign: Thanks for sending in the PR! Your idea sounds cool, but I feel that this might be too specific a use-case to incorporate in this library. If you haven't done that already, I'd encourage you to fork this library and change the API as you see fit.

Hope that makes sense, and I'm sorry that I couldn't merge your awesome PR. Thanks again 👍

stahlmanDesign:
Thanks, I understand

PR #103, opened 2016-08-23
Split clipboard data by any whitespace characters and commas

johnste:


johnste:
Please let me know if this doesn't make sense. Maybe the regex could be configurable? This would support my use case (posting multiple lines of data).

lib/ReactTags.js
```diff
@@ -190,7 +190,7 @@ var ReactTags = React.createClass({
     e.preventDefault()
     const clipboardData = e.clipboardData || window.clipboardData
     const string = clipboardData.getData('text')
-    string.split(/[ ,]+/).forEach((tag) => this.props.handleAddition(tag))
+    string.split(/[\s,]+/).forEach((tag) => this.props.handleAddition(tag))
```
prakhar1989: Can you add a comment on what this regex supposed to do? Will help developers save a Google search :)

test/reactTags-test.js
```diff
@@ -56,6 +56,29 @@ describe("ReactTags", () => {
 
   });
 
+  it("handles the paste event and splits the clipboard on whitespace or comma characters", () => {
```
prakhar1989: Tests looks sweet! 🍬  Thanks for adding the tests 🍻

prakhar1989:
Heya @johnste - This looks great! Thanks for sending this PR! I think this is a reasonable default regex. If more users want this to be configurable we can do that later on. For now, I added just one comment for you to address following which I can merge and ship this change. 

Thanks again for sending this over 👍

lib/ReactTags.js
johnste: Sure!

prakhar1989:
@johnste : Merged and published your changes! Thanks again for contributing 👯

johnste:
My pleasure - thanks for a good react component.

Issue #105, opened 2016-08-30
handleInputBlur function is not working


wendylw:
Bind myself function on handleInputBlur, but can not working

prakhar1989:
Hi @wendylw, thanks for opening the issue. If you look at the code for [`handleInputBlur`](https://github.com/prakhar1989/react-tags/blob/9de77b8b9c823a7c498168c4fba6f5db4d34d44d/lib/ReactTags.js#L120-L126), you'll see that the handler is triggered **only when** the textfield is not blank. Did you try out this scenario too?

I've created a webpackbin example for you to see how this works. You can try this here: http://www.webpackbin.com/4yzaRtA9b (click on "Log" button in the menubar to see the output)

wendylw:
Thank you for solution my problem

ntsim:
Hi @prakhar1989, I was playing around with your component today (good work btw) and am surprised by the above `onBlur` behaviour. I'm curious why it shouldn't call the passed `handleOnBlur` if the field is empty?

prakhar1989:
Hi @ntsim, interesting question. I had to check the history to answer that, and it seems the answer lies in [this issue](https://github.com/prakhar1989/react-tags/issues/84) by @Wildhoney. 

The idea behind this change was to automatically convert the text in the input field in case the user tabs over and hence it makes sense to trigger in when the field is non-empty.

Thinking about it now though, I feel the event should always be triggered irrespective of whether the field is empty or not so that it's consistent with what developers expect. The overriding function can then accordingly deal with the empty value scenario. What do you think?

ntsim:
Sounds like a great idea. A validation check in the passed `handleInputBlur` function would be trivial for a developer to implement, but having such a check baked into the API would considerably reduce the amount of control they would have over the behaviour. Generally fewer assumptions are better for building generic components :+1:

Wildhoney:
Will open a PR for the modified behaviour 👍

PR #109, opened 2016-09-12
Invoking handleInputBlur irrespective of input value

Wildhoney:
#105


prakhar1989:
This is great! Thanks Adam for picking this up =)

PR #114, opened 2016-09-27
Allow setting the active suggestion className

andrefgneves:


prakhar1989:
Hey, sorry for the infinitely late follow up. Since it wouldn't be fair to ask you to update the PR, I've gone ahead and added the className that you were looking for in #159.

Sorry again for not responding back.

PR #116, opened 2016-10-13
Fix adding a new tag in SelectionMode

klacointe:
Step to reproduce:
- Run the `npm run dev` example with `autocomplete={true}` props
- Type `Ind` and navigate in the suggestions without validate
- Type additional characters to have no suggestions, ie `Indy`
- Press enter
- The new tag is not added

=> `Uncaught TypeError: Cannot read property 'toLowerCase' of undefined`

### DIFF
--- a/lib/ReactTags.js
+++ b/lib/ReactTags.js
@@ -147,7 +147,7 @@ var ReactTags = React.createClass({
       }
 
       if (query !== "") {
-        if (this.state.selectionMode) {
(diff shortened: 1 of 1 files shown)


klacointe:
@prakhar1989 up?

prakhar1989:
Heya @klacointe, so sorry - for some reason I completely missed this PR! I just reproduced the bug and your fix seems great. 

Thanks for finding this and sending in the fix.

klacointe:
Thanks!

PR #117, opened 2016-10-20
Add the ability to pass in a custom className for the <input> element

justbaker:
In regards to issue #32 , adds the ability to pass in a custom className for the `<input>` element, rather than just the outer `<div>`.


prakhar1989:
Sorry for the late follow up - this looks awesome! Merging it now.

Since I'm away from my main computer, I'll be releasing it later in the day.

Thanks a ton for contributing!

PR #121, opened 2016-11-05
Add new prop inputType to allow set input type

KhueHuynh:
Hi Prakhar,

I just add input type prop. I think it's useful. I just ran build command, some build file is changed without my hand. 
Please help me review file lib/ReactTags.js

Thanks,Khue


mrdannael:
Rebase with master and do not commit `dist/` and `dist-modules/` files.

Issue #128, opened 2017-01-06
Focus when removing tags


Wildhoney:
Expectedly, when removing a tag the focus is transferred from the `input` to the close button. However, it'd be nice to optionally transfer the focus **back** to the `input` immediately, thus improving UX.

Any thoughts on the best way to implement something like this? Would it require a lib change?

prakhar1989:
Hey Adam, not sure I understand you correctly. When I use my mouse to click on the `x` button, the focus doesn't change, but when I used <kbd>backspace</kbd>, I see that the focus remains on the input. Are you referring to something else?

Wildhoney:
For me in Chrome 55.0.2883.95:

- [Navigate to the demo](http://www.webpackbin.com/EJdPySdLG);
- Focus on the `input` (by default);
- Enter a country
- Delete the country

**Expected:** Focus remains on the `input` node
**Actual:** Focus is transferred to the `x` and thus you need to click on the `input` again

Wildhoney:
![Focus](http://i.imgur.com/gyzdm2O.gif)

prakhar1989:
Hey Adam, apologies for the late update - i just released a fix for this. Feel free to reopen the issue if it still persists.

Issue #112, opened 2016-09-21
Not able to create a new tag after deleting the existing tags


ukils:
steps followed..
1. Created some tags.
2. type a new tag , don't press separator.
3. delete all the existing tags.
4. now try to press separator.
This does not create a new tag unless you modify what you have typed.

mrdannael:
fix already on master

PR #135, opened 2017-02-01
Fix to focus input and not to reset query state after deleting tags #128 #112

ming-soon:


mrdannael:
Rebase with master.

mrdannael:
fixed on master

PR #143, opened 2017-03-18
allow disabling tag addition when a value was pasted into input field

sverdlovka:
I wanted to have ability to control whether tags should be added automatically when pasting values to the input field, but there was no such functionality. New property `allowAdditionFromPaste` is `true` by default, which doesn't change the default behaviour.


mrdannael:
Rebase with master.

mrdannael:
new prop added to master

sverdlovka:
awesome, thanks!

PR #147, opened 2017-03-22
fix re-rendering of Suggestion to make use of nextProps

cklab:
Suggestions don't always render when they're supposed to (e.g. hitting `minQueryLength` is not sufficient, always have to hit `minQueryLength+1` for a render to occur). Looking at the implementation of `shouldComponentUpdate` and `shouldRenderSuggestions` suggests two issues

1. `shouldComponentUpdate` is ignoring nextProps
2. `shouldRenderSuggestions` is ignoring the `query` argument

### DIFF
--- a/lib/Suggestions.js
+++ b/lib/Suggestions.js
@@ -33,8 +33,10 @@ class Suggestions extends Component {
     const shouldRenderSuggestions =
       props.shouldRenderSuggestions || this.shouldRenderSuggestions;
     return (
-      !isEqual(this.props.suggestions, nextProps.suggestions) ||
-      shouldRenderSuggestions(props.query)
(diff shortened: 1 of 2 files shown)


cklab:
Rebased

prakhar1989:
@cklab - Thanks a lot for rebasing! I recently adding a [CONTRIBUTION](https://github.com/prakhar1989/react-tags/blob/master/CONTRIBUTING.md) file. Can you please update the PR accordingly? Thanks!

cklab:
Ah, nice! I always wondered if checking in the dist* was desirable. Done & done

prakhar1989:
Looks great! Thanks for contributing @cklab!

PR #149, opened 2017-04-13
Enhancement on flexibility of label rendering

songyanho:
This patch allow custom rendering of label rather than only showing one of the value.

I have added one example that I am using now to demonstrate the usage.


songyanho:
I try to build and the example is not working due to an error of t.getContext(), but I managed to use it flawlessly with my project. Maybe there is some problem with webpack configuration or React?

teldosas:
@songyanho I had the same issue. Addressed with #151

PR #151, opened 2017-04-15
Update react-dnd to fix 'getContext not a function'

teldosas:
After building the example I get a `t.getContext` is not a function error. After investigating turns out it was `example/vendor/ReactDND.min.js` to blame. Updating it from npmjs registry fixed the issue.


prakhar1989:
I merged these changes, and I'm still getting `t.getContext` is not a function. Are you seeing the same?

teldosas:
After running `npm run build` and `npm run dev`, no I don't get the error anymore. Hot code replacement doesn't work in `npm run dev` don't know why.

prakhar1989:
Yup, seems to be working for me now too! Weird. Let me publish a new version with these changes.

PR #152, opened 2017-04-15
Fix tests

teldosas:


teldosas:
@prakhar1989 I found what was failing the builds

prakhar1989:
This is AWESOME @teldosas! Thank you so much for sending this fix!

PR #154, opened 2017-04-15
Add webpack configuration that works

teldosas:


prakhar1989:
This looks fantastic! Thank you @teldosas! I didn't get a chance to dive in and keep up with the idioms in webpack. Thank you so much for sending in these fixes!

PR #158, opened 2017-04-16
Tests for focus in ReactTags and adding input focus on delete. Fixes #128

### LINKED ISSUE #128: Focus when removing tags

prakhar1989:


test/reactTags.test.js
```diff
@@ -1,155 +1,189 @@
...
-  });
+test("focus on input by default", () => {
+  const $el = mount(mockItem());
+  expect(document.activeElement.tagName).to.equal("INPUT");
+  expect(document.activeElement.className).to.equal("ReactTags__tagInputField");
+});
+
+test("should not focus on input if autofocus is false", () => {
+  const $el = mount(mockItem({autofocus: false}));
+  expect(document.activeElement.tagName).to.equal("BODY");
```
prakhar1989: This is failing - no idea why 😞

prakhar1989: @rnicholus: Would you happen to have any experience with Jest? I'm having a very hard time figuring out why this test is failing? Any tips?

rnicholus: Is it possible that the previous test is affecting this one? In other words, is the focused `input` from the previous test still in the DOM?

prakhar1989: Yup, my guess is on that one too. Changing the order of tests fixes the problem. Is there a way to render a test separately in Jest?

rnicholus: I think what you probably want to do instead is to ensure ReactTags is unmounted after every test. If that doesn't work for some reason, let me know and I'll take a closer look when I get a block of time this week.

prakhar1989: Just tried unmounting and it works! Thanks for the help, @rnicholus. Do you think this an issue with Jest or Enzyme? I can't find a mention of this on Enzyme docs at least.

rnicholus: I think it is not an issue with either of these. Normally it may not be an issue to have multiple ReactTags components mounted in a test environment. But since you are checking state that exists outside of the component and is common to all mounted components (`document`), then you will need to be careful to cleanup after your tests by unmounting. I don't think there is an expectation that enzyme will auto unmount after each test, or that jest or enzyme will cleanup the DOM after each test.

prakhar1989: Ah, valid point. Thanks a ton for jumping on this and helping me figure this out, Ray! Really appreciate it :)

rnicholus: No problem. Glad I could be of assistance.

rnicholus: One suggestion: you may want to consider unmounting `$el` in an `afterEach` block in your test spec instead. This should prevent the problem from returning if the tests are, for some reason, not run in the declared order, or from a refactor that reorders the tests or inserts a new one in-between the two.

prakhar1989: Sound good! Will do the needful - thanks for the review :)

PR #166, opened 2017-04-20
Fix reset input when autofocus=false

ZachGawlik:
Fixes https://github.com/prakhar1989/react-tags/issues/165

Before bug, that can be replicated by setting `autofocus={false}` on the example.
bug gif

The bug came from https://github.com/prakhar1989/react-tags/commit/4429be7922173e25a764fb26ced8898a15e49edf?diff=split#diff-833c370f0b7e67628628a041d1984df3L211 , where after tag addition the input would only reset if autofocus wasn't false. I decided to move the if condition into only `componentDidMountt`, as I am under the assumption that the input should be focused after tag deletion regardless of the autofocus prop. 

I didn't see a way to add in a test to ensure the input gets cleared after adding a tag, but if that is possible please point me in the right direction!

### DIFF
--- a/lib/ReactTags.js
+++ b/lib/ReactTags.js
@@ -57,15 +57,15 @@ class ReactTags extends Component {
   }
 
   resetAndFocusInput() {
-    const { autofocus, readOnly } = this.props;
-    if (autofocus && !readOnly) {
-      this.textInput.value = "";
-      this.textInput.focus();
-    }
+    this.textInput.value = "";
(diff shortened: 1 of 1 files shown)


StefanFeser:
Any idea when this will be merged? It's a quite annoying bug for any user of my application.

prakhar1989:
@ZachGawlik Thanks for sending this PR! Releasing this in a new version right away!

PR #181, opened 2017-07-26
Change Suggestions to use a ref function.

rbscott:
The newer React standard suggests using a function
for ref instead of a string. This makes a small
tweak to change Suggestions to this pattern.

The main change is pretty small and I believe it is already tested by an existing test. Let me know if that isn't the case and I will add one. The bulk of the changes came from `npm run format` as suggested in the coding guidelines and included the output.


mrdannael:
Already on master. :smiley:

PR #193, opened 2017-09-04
Fix readme example

StefanFeser:
Hi,

Since I just added your component to a new project I copied the example from your README page to get started. 
I found two problems which threw errors so I thought I quickly fix it.

```
Line 12:  'Countries' is not defined
```

and 

```
Syntax error: Unexpected token:  });
```

Greets

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ class App extends React.Component {
             </div>
         )
     }
-});
+};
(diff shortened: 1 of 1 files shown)


mrdannael:
Thanks for the PR. Merged. :beers:

PR #196, opened 2017-09-07
Remove key code

AndrewFinlay:
Event.keyCode is deprecated, use event.key instead.


lib/ReactTags.js
```diff
@@ -127,7 +127,7 @@ class ReactTags extends Component {
     let { query, selectedIndex, suggestions } = this.state;
 
     // hide suggestions menu on escape
-    if (e.keyCode === Keys.ESCAPE) {
+    if (e.key === "Escape") {
```
ad1992: @AndrewFinlay can you update the constant `Keys` declared in this file with these values instead of using hard coded values in this file?

package.json
```diff
@@ -50,7 +50,7 @@
     "chai": "^4.0.1",
     "enzyme": "^2.3.0 || ^3.0.0",
     "enzyme-adapter-react-16": "^1.0.0",
-    "jest": "^20.0.0",
+    "jest": "^21.0.1",
```
ad1992: can be locked at exact version `21.0.1`?

lib/ReactTags.js
AndrewFinlay: Will do

AndrewFinlay: Looking at it more closely, I think that this breaks props.delimiters.

package.json
AndrewFinlay: That's up to you, if it's locked then you can be sure that jest version differences aren't causing problems in the tests.

ad1992: Sure I would recommend locking it at exact version.

ad1992:
@AndrewFinlay can you change your base in the PR to master and resolve the conflicts if any.

AndrewFinlay:
I've done the rebase but it doesn't look quite right to me, if you could take a look and if it's good then great.  Otherwise let me know what needs to be done I'll see what I can do

ad1992:
@AndrewFinlay since this is an important change which will affect the whole code base so before I merge this branch to master , there are some pending changes which I am working on master branch so I will let you know once master is ready so you will have to rebase again that time so don't rebase now. I will let you know when to rebase.

AndrewFinlay:
No worries, but we should move the conversation to the new PR I've added with the working rebase.  It's a better point to continue the discussion then we can abandon and close this one

ad1992:
Sure @AndrewFinlay. Closing this PR as new PR has been created https://github.com/prakhar1989/react-tags/pull/260

PR #207, opened 2017-10-25
Allow wrapping Tag via tagHoc prop (eg for tags that are links)

helixbass:
For our use case, wanted each tag to be a link to a listing page of things tagged with that tag

My solution was to allow passing a higher-order component to wrap `Tag` via the `tagHoc` prop


mrdannael:
Rebase with master and do not commit `dist/` and `dist-modules/` files.

PR #209, opened 2017-11-03
Update to use react 16

glaubinix:
Update to react 16
* `inputValue` props is necessary because setting the input field value throws `TypeError: Can't add property value, object is not extensible` see the test that was split into 2


prakhar1989:
Thank you!!! 🚀

PR #214, opened 2017-11-28
support minQueryLength 0 value

### LINKED ISSUE #213: does react-tag support listing suggestion when minQueryLength={0}?
does react-tag support listing suggestion when `minQueryLength={0}`? I want a list of suggestions to show even if I haven't typed anything yet. I tried `minQueryLength` but it does not work

xxryan1234:
- closes #213

### DIFF
--- a/lib/Suggestions.js
+++ b/lib/Suggestions.js
@@ -64,7 +64,10 @@ class Suggestions extends Component {
 
   shouldRenderSuggestions = query => {
     const { props } = this;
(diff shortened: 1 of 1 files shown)


gregnavis:
It seems this will show suggestions even if the user hasn't clicked the element yet. This may be confusing especially if there are multiple elements on a single page. I don't think they all should be expanded by default.

mrdannael:
What about [this](https://github.com/prakhar1989/react-tags/pull/210) PR?

PR #216, opened 2017-12-04
Fix propTypes definition

jacobp100:
I get the error `Warning: Component ReactTags declared `PropTypes` instead of `propTypes`. Did you misspell the property assignment?`. This fixes that (and adds prop-types validation)


jacobp100:
Closing in favour of #219

PR #236, opened 2018-02-24
Add stackblitz for demoing library usage

prakhar1989:
I think the example included in the repo is hard to maintain and doesn't quite show the actual usage of the library, so adding a stackblitz demo which users can fork and play right in the browser might be a nice way to experiment.

Here's the demo: https://react-tag-input.stackblitz.io

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ Because I was looking for an excuse to build a standalone component and publish 
 ### Demo
 ![img](demo.gif)
 
-Check it out [here](http://prakhar1989.github.io/react-tags/example/)
+Check it out [here](https://react-tag-input.stackblitz.io)
 
 ### Installation
 The preferred way of using the component is via NPM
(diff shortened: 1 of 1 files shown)


mrdannael:
Awesome! Can we also add to README.md a direct link to editor with react-tags?
`https://stackblitz.com/edit/react-tag-input`

README.md
```diff
@@ -24,15 +24,15 @@ Because I was looking for an excuse to build a standalone component and publish
...
+Check it out [here](https://react-tag-input.stackblitz.io)
 
 ### Installation
 The preferred way of using the component is via NPM
 
 ```
 npm install --save react-tag-input
 ```
-It is, however, also available to be used separately (`dist/ReactTags.min.js`). If you prefer this method remember to include [ReactDND](https://github.com/gaearon/react-dnd) as a dependancy. Refer to the [demo](http://prakhar1989.github.io/react-tags/example/) to see how this works.
+It is, however, also available to be used separately (`dist/ReactTags.min.js`). If you prefer this method remember to include [ReactDND](https://github.com/gaearon/react-dnd) as a dependancy. Refer to the [example](https://stackblitz.com/edit/react-tag-input) to see how this works.
```
prakhar1989: Added the direct link to the editor here. Should we add it someplace as well or does this suffice?

mrdannael: Should be enough :+1:

PR #245, opened 2018-03-07
fix #235, don't allow duplicate tags

### LINKED ISSUE #235: Duplicate tags | You can enter the same tag twice

ad1992:
@mrdannael @prakhar1989  please look into this.

### DIFF
--- a/lib/ReactTags.js
+++ b/lib/ReactTags.js
@@ -137,7 +137,7 @@ class ReactTags extends Component {
   handleFocus(e) {
     const value = e.target.value.trim();
     if (this.props.handleInputFocus) {
-        this.props.handleInputFocus(value);
+      this.props.handleInputFocus(value);
     }
     this.setState({ isFocused: true });
   }
--- a/lib/ReactTags.js
+++ b/lib/ReactTags.js
@@ -95,9 +95,9 @@ class ReactTags extends Component {
   handleDelete(i, e) {
     this.props.handleDelete(i, e);
     if (!this.props.resetInputOnDelete) {
-        this.textInput.focus();
+      this.textInput.focus();
     } else {
-        this.resetAndFocusInput();
+      this.resetAndFocusInput();
     }
     e.stopPropagation();
   }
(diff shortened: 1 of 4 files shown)


lib/ReactTags.js
```diff
@@ -95,9 +95,9 @@ class ReactTags extends Component {
   handleDelete(i, e) {
     this.props.handleDelete(i, e);
     if (!this.props.resetInputOnDelete) {
-        this.textInput.focus();
```
prakhar1989: Would recommend running this through prettier `npm run format` to make sure we are consistent formatting / linting wise.

lib/ReactTags.js
```diff
@@ -244,6 +244,13 @@ class ReactTags extends Component {
   }
 
   addTag(tag) {
+    const { tags } = this.props;
+    const existingTags = tags.map(tag => tag.text.toLowerCase());
+
+    // Return if tag has been already added
+    if (existingTags.indexOf(tag.toLowerCase()) >= 0) {
```
prakhar1989: How about using includes? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

test/reactTags.test.js
```diff
@@ -155,6 +153,22 @@ test("handles the paste event and splits the clipboard on delimiters", () => {
...
 
+test("should not allow duplicate tags", () => {
+  const actual = [];
+  const $el = mount(
+    mockItem({
+      handleAddition(tag) {
+        actual.push(tag);
+      },
+    })
+  );
```
prakhar1989: Can you add an assertion for checking if 'Apple' is already there?

lib/ReactTags.js
ad1992: @prakhar1989 since includes doesn't support IE thats why I used indexOf

lib/ReactTags.js
ad1992: I ran `npm run format` which resulted in this change. Might be  `npm run format ` was not run when the above code was pushed.

test/reactTags.test.js
ad1992: added 👍

prakhar1989:
Looks great, thanks for making this change Aakansha 👍 

LGTM from my side, but I'd recommend waiting for @mrdannael to take a quick look.

mrdannael:
Looks good to me :)

ad1992:
@prakhar1989 @mrdannael can this pr be merged?

mrdannael:
Merged :+1:

ad1992:
@mrdannael can you also update the link http://prakhar1989.github.io/react-tags/example/ so it does't add duplicate tags

mrdannael:
I'll fix it with another issue.

PR #255, opened 2018-03-15
Add document for inline attribute

hoantran-it:
Firstly, I want to say thank you for this library! It's very helpful!
When I want to scroll selected tags only, I dive deeply inside source code and find out that we already support it, just missing some document. I create this PR to contribute a bit for project.


README.md
```diff
@@ -365,6 +366,34 @@ The maxLength attribute added to the input. Specifies the maximum number of char
     ...>
 ```
 
+<a name="inline"></a>
+##### inline (optional)
+The inline attribute use to determine that selected tags and input field will be put in the same div block or not. If you want to separate it and apply auto scroll for just selected tags only, change this attribute. Specifies the inline attribute.
```
ad1992: The inline attribute is deciding whether to render the  input field in same div as selected tags or the sibling div. So I think we can remove this line
`If you want to separate it and apply auto scroll for just selected tags only, change this attribute. Specifies the inline attribute
`

ad1992: Can you change the text to  `The inline attributes decides whether the input fields and selected tags will be rendered in-line` ?

README.md
```diff
@@ -365,6 +366,34 @@ The maxLength attribute added to the input. Specifies the maximum number of char
...
+    ...>
+```
+
+If inline is true (default value)
+
+```
+<div className="...">
+	{tagItems}
+	{tagInput}
+</div>
```
ad1992: I would recommend adding screen shots in both cases instead of showing how its rendered in div

hoantran-it: Hi @ad1992, I fix them all as request. Please help to take a look. Thanks!

ad1992: Thanks @hoantran-it  :) .The screenshots look too big and in both the case it seems that they are not inline. Can you decrease the size of screenshots may be something like this 
screen shot 2018-03-17 at 5 20 12 am
screen shot 2018-03-17 at 5 21 34 am

README.md
```diff
@@ -365,6 +366,38 @@ The maxLength attribute added to the input. Specifies the maximum number of char
...
+<ReactTags
+    inline = {false}
+    ...>
+```
+
+If inline is true (default value)
+
+```
+<div className="...">
+	{tagItems}
```
ad1992: This won't be needed after adding the screenshot :)
Can you do it the below way
```
<ReactTags
  inline
>
```
add the screenshot when inline true

```
<ReactTags
  inline={false}
>
```
Add the screenshot for the inline false

README.md
```diff
@@ -365,6 +366,38 @@ The maxLength attribute added to the input. Specifies the maximum number of char
...
+
+![img](docs/inline-true.png)
+
+If inline is false
+
+```
+<div className="...">
+	{tagItems}
+</div>
+{tagInput}
```
ad1992: this can also be removed.

README.md
hoantran-it: Hi @ad1992 can I use your screenshot, i think it's suitable :)

ad1992: sure 👍 @hoantran-it

hoantran-it: @ad1992 Thanks! Updated, please help to take a look :)

ad1992:
looks good @hoantran-it :)

Issue #251, opened 2018-03-12
iOS can not click on item in suggestion list


tzehe:
Hi, 

I am experiencing a strange behaviour after clicking on a suggestion item from my mobile device. The item is not added to the tags. I also tested this on the example page and had the same result.
After this I add `onTouchStart={props.handleClick.bind(null, i)}` in Suggestions.js:

`<li
            key={i}
            onMouseDown={props.handleClick.bind(null, i)}
            onMouseOver={props.handleHover.bind(null, i)}
            onTouchStart={props.handleClick.bind(null, i)}
            className={
              i == props.selectedIndex ? props.classNames.activeSuggestion : ""
            }>
            <span dangerouslySetInnerHTML={this.markIt(item, props.query)} />
          </li>`

Test Devices: 

iPhone 5 and X

The onTouchStart event is equivalent to onMouseOver but for mobile devices. This fixed the issue for me. 

Are you also experiencing this issue? Maybe this should be included?

ad1992:
@tzehe would you like to give a PR for this fix?

tzehe:
Yes I will give a PR later.

PR #261, opened 2018-03-16
add touch start event to Suggestion component

tzehe:
The reason for this is to guarantee that handleClick is also called when suggestion item is 'touched' by user.
Corresponding issue: iOS can not click on item in suggestion list #251

### DIFF
--- a/lib/Suggestions.js
+++ b/lib/Suggestions.js
@@ -86,6 +86,7 @@ class Suggestions extends Component {
             key={i}
             onMouseDown={props.handleClick.bind(null, i)}
             onMouseOver={props.handleHover.bind(null, i)}
+            onTouchStart={props.handleClick.bind(null, i)}
             className={
               i == props.selectedIndex ? props.classNames.activeSuggestion : ""
             }>
--- a/test/suggestions.test.js
+++ b/test/suggestions.test.js
@@ -3,6 +3,7 @@ import ReactDOM from "react-dom";
 import { expect } from "chai";
 import { shallow, mount, render } from "enzyme";
 import { spy } from "sinon";
+import sinon from "sinon";
 import Suggestions from "../lib/Suggestions";
 import noop from "lodash/noop";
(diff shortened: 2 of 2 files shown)


lib/Suggestions.js
```diff
@@ -86,6 +86,7 @@ class Suggestions extends Component {
             key={i}
             onMouseDown={props.handleClick.bind(null, i)}
             onMouseOver={props.handleHover.bind(null, i)}
+            onTouchStart={props.handleClick.bind(null, i)}
```
ad1992: this doesn't have good browser compatibility https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/ontouchstart
cc @mrdannael @prakhar1989 what do you suggest ?

lib/Suggestions.js
```diff
@@ -114,4 +115,4 @@ class Suggestions extends Component {
   }
 }
 
-export default Suggestions;
+export default Suggestions;
```
ad1992: can you add a new line

test/suggestions.test.js
```diff
@@ -155,4 +156,14 @@ describe("Suggestions", function() {
...
+  test("should call handleClick when touched", function() {
+    const s = spy();
+    const $el = mount(mockItem({ handleClick: s }));
+    $el
+      .find("li")
+      .at(0)
+      .simulate("touchStart");
+    expect(s.calledOnce).to.be.true;
+  });
+});
```
ad1992: can you add a new line

test/suggestions.test.js
```diff
@@ -155,4 +156,14 @@ describe("Suggestions", function() {
 
     expect($el.componentDidUpdate.called).to.equal(true);
   });
-});
+
+  test("should call handleClick when touched", function() {
+    const s = spy();
```
ad1992: spyHandleClick

PR #275, opened 2018-03-22
fix prettier

ad1992:


prakhar1989:
Awesome! Thanks for doing this! :tada:

PR #289, opened 2018-04-08
add sanity checks in release.sh

prakhar1989:


release.sh
```diff
@@ -1,11 +1,29 @@
...
-git push origin --tags
-git commit -am "upgrading to version $1.$2.$3"
-git push origin master
-npm publish
-echo "Version published successfully"
+
+# releases a new version
+release () {
+  echo "Version $1 in progress"
+  git checkout master
```
ad1992: should we use `&&` for safe check for all the commands here so that next command runs only if prev command was successful ?

Issue #194, opened 2017-09-06
Add option to disable input.


kingllama:
There's a `readOnly` property right now, but it would be great to be able to `disable` the input.

sjsn:
+1 I need closable tags without the input.

ad1992:
@kingllama would you give a PR for this feature ?

kingllama:
@ad1992 No promises but I can definitely take a look this weekend. My react skills and free time have certainly improved since I opened the issue.

ad1992:
@sjsn I am sorry I didn't get you what you need. Can you please explain ?

sjsn:
@ad1992 The pull request kingllama made addresses what I needed. Thanks!

dmossie:
@ad1992 Was this ever implemented? This would be a great feature to have

ad1992:
@dmossie there was a PR but the pr was not merged as some fixes were left in it.  Here is the PR https://github.com/prakhar1989/react-tags/pull/296

atiqueansari1987:
@ad1992 I will try and submit PR this weekend.

ad1992:
closed via #726, this will be available in next version 🎉

PR #296, opened 2018-04-14
Added option to disable the input

kingllama:
Issue #194

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,7 @@ Otherwise, you can simply import along with the backend itself (as shown above).
 - [`minQueryLength`](#minQueryLength)
 - [`removeComponent`](#removeComponent)
 - [`autocomplete`](#autocomplete)
+- [`disabled`](#disabled)
 - [`readOnly`](#readOnly)
 - [`name`](#nameOption)
 - [`id`](#idOption)
@@ -349,6 +350,10 @@ Useful for enhancing data entry workflows for your users by ensuring the first m
 - `false` (default) - tags are not autocompleted on enter/delimiter
 
 This option has no effect if there are no [`suggestions`](#suggestionsOption).
+
+<a name="disabled"></a>
+##### disabled (optional)
+Disables the input box, preventing text from being entered.
 
 <a name="readOnly"></a>
 ##### readOnly (optional)
--- a/lib/ReactTags.js
+++ b/lib/ReactTags.js
@@ -346,6 +346,7 @@ class ReactTags extends Component {
           id={inputId}
           maxLength={maxLength}
           value={this.props.inputValue}
+          disabled={this.props.disabled}
         />
 
         <Suggestions
@@ -377,10 +378,12 @@ class ReactTags extends Component {
 ReactTags.propTypes = {
   placeholder: PropTypes.string,
   labelField: PropTypes.string,
-  suggestions: PropTypes.arrayOf(PropTypes.shape({
-    id: PropTypes.string.isRequired,
-    text: PropTypes.string.isRequired
-  })),
+  suggestions: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.string.isRequired,
+      text: PropTypes.string.isRequired,
+    })
+  ),
   delimiters: PropTypes.array,
   autofocus: PropTypes.bool,
   inline: PropTypes.bool,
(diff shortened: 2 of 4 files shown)


lib/ReactTags.js
```diff
@@ -399,16 +402,19 @@ ReactTags.propTypes = {
   shouldRenderSuggestions: PropTypes.func,
   removeComponent: PropTypes.func,
   autocomplete: PropTypes.oneOfType([PropTypes.bool, PropTypes.number]),
+  disabled: PropTypes.bool,
```
ad1992: can you `disabled` false in defaultProps ?

kingllama:
> Since the input is disabled then should we show the input field ?
Wha do you think ?

This is doable, but I think the better solution is to just leave the field disabled. From there, users can hide it on their own with some CSS, such as `ReactTags__tagInputField:disabled { display: none }`

The use case here would be when the user wants to be able to drag and drop the tags or remove the tags but prevent any input. Furthermore, as of now you can effectively block the UI for drag and drop (by not passing in a `handleDrag` function) and delete (again by not passing in a `handleDelete`), but there is no easy way to block the input UI without disabling the entire component through `readOnly`.

My specific scenario in this case was that we were doing conditional validation, and for consistency, wanted the entire field to "grey out". That way it would be obvious to the end user that they are not able to edit the feild at this time, but it is something that can be editable and not just a list of boxes.

ad1992:
@kingllama can you also change the placeholder to `The input field is disabled` by default in case `disabled` is true and `placeholder` props is not passed ? And also add a test for the same.

ad1992:
@kingllama any updates?

lib/ReactTags.js
```diff
@@ -30,6 +35,7 @@ class ReactTags extends Component {
     this.resetAndFocusInput = this.resetAndFocusInput.bind(this);
     this.handleSuggestionHover = this.handleSuggestionHover.bind(this);
     this.handleSuggestionClick = this.handleSuggestionClick.bind(this);
+    this.getPlaceholderText = this.getPlaceholderText.bind(this);
```
ad1992: no need to bind

lib/ReactTags.js
```diff
@@ -63,6 +69,18 @@ class ReactTags extends Component {
     });
   }
 
+  getPlaceholderText() {
+    if (this.props.placeholder !== undefined) {
```
ad1992:
```js
const { placeholder, disabled } = this.props;
    return placeholder
      ? placeholder
      : disabled ? DISABLED_PLACEHOLDER : DEFAULT_PLACEHOLDER;
```

test/reactTags.test.js
```diff
@@ -245,6 +245,29 @@ describe('Test ReactTags', () => {
     $input.simulate('keyDown', { keyCode: DOWN_ARROW_KEY_CODE });
   });
 
+  test('disables the input with disabled prop', () => {
```
ad1992: `should disable the input when disabled prop is true`

ad1992: can you actually put all the tests with the `disabled` prop inside a `describe` block ?
`describe('tests disabled prop')`

test/reactTags.test.js
```diff
@@ -245,6 +245,29 @@ describe('Test ReactTags', () => {
...
+        disabled: true,
+      })
+    );
+
+    const $input = $el.find('.ReactTags__tagInputField');
+
+    expect($input.prop('disabled')).to.be.true;
+  });
+
+  test('default placeholder text is different when disabled', () => {
```
ad1992: `should show DISABLED_PLACEHOLDER when input field is disabled`

test/reactTags.test.js
```diff
@@ -245,6 +245,29 @@ describe('Test ReactTags', () => {
...
+    const $el = mount(
+      mockItem({
+        disabled: true,
+      })
+    );
+
+    const $input = $el.find('.ReactTags__tagInputField');
+    expect($input.prop('placeholder')).to.equal('This input field is disabled');
```
ad1992: import `constants.js` here and use it directly

test/reactTags.test.js
```diff
@@ -245,6 +245,29 @@ describe('Test ReactTags', () => {
...
+  });
+
```
ad1992: Can you also add a test inside the describe block for `disabled`
* when disabled prop is false, default placeholder is shown
* when placeholder prop is passed, placeholder prop is shown

README.md
```diff
@@ -350,6 +351,10 @@ Useful for enhancing data entry workflows for your users by ensuring the first m
 
 This option has no effect if there are no [`suggestions`](#suggestionsOption).
 
+<a name="disabled"></a>
+##### disabled (optional)
+Disables the input box, preventing text from being entered.
```
ad1992: `Disables the input field, preventing text from being entered.`

PR #302, opened 2018-04-25
redundant {} in Usage state code

ithinco:


ad1992:
Thanks for fixing the issue @ithinco

PR #304, opened 2018-04-26
Give handleAddition and handleDelete default props

### LINKED ISSUE #303: `handleAddition` and `handleDelete` should not be required props
Given that `handleAddition` and `handleDelete` aren't used or accessible when `readOnly` is `true`, it seems like having a default noop function should suffice for both and the `.isRequired` can be removed.

jpbufe3:
When `readOnly` is `true`, both `handleAddtion` and `handleDelete` are unused. Provide a noop function as a default prop and remove the requirement for them to be provided.

Fixes #303

### DIFF
--- a/lib/ReactTags.js
+++ b/lib/ReactTags.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { DragDropContext } from 'react-dnd';
 import HTML5Backend from 'react-dnd-html5-backend';
-import noop from 'lodash/noop';
+import { noop } from 'lodash';
 import Suggestions from './Suggestions';
 import PropTypes from 'prop-types';
 import Tag from './Tag';
--- a/lib/ReactTags.js
+++ b/lib/ReactTags.js
@@ -384,8 +384,8 @@ ReactTags.propTypes = {
   delimiters: PropTypes.array,
   autofocus: PropTypes.bool,
   inline: PropTypes.bool,
-  handleDelete: PropTypes.func.isRequired,
-  handleAddition: PropTypes.func.isRequired,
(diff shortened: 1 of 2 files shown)


lib/ReactTags.js
```diff
@@ -384,8 +384,8 @@ ReactTags.propTypes = {
   delimiters: PropTypes.array,
   autofocus: PropTypes.bool,
   inline: PropTypes.bool,
-  handleDelete: PropTypes.func.isRequired,
-  handleAddition: PropTypes.func.isRequired,
+  handleDelete: PropTypes.func,
```
ad1992: can you move proptypes and default props inside the class and make it static ?

jpbufe3: I certainly can, but that seems out of scope for this PR. if you'd like me to, I can do it, but I can also put up a separate PR that makes the mechanical change to separate it from the functionality change. up to you

ad1992: yes that works. You can give a separate PR for the change.

ad1992:
@jpbufe3 since you made these props optional can you also remove them from tests as its passed as `noop` which is now taken care in the component itslef

lib/ReactTags.js
```diff
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
-import ReactDOM from 'react-dom';
 import { DragDropContext } from 'react-dnd';
 import HTML5Backend from 'react-dnd-html5-backend';
+import noop from 'lodash/noop';
```
ad1992: one last change :) `import { noop } from 'lodash'`

PR #306, opened 2018-04-26
Fix pasting into a tag input

### LINKED ISSUE #305: Paste calls `addTag` with the wrong type
Pasting currently splits the string into an array of strings and then passes each one directly to `addTag`. However, `addTag` now expects an object with `id` and `text`.

See: https://github.com/prakhar1989/react-tags/blob/c59ff527f08b8f307f69d883be8bfe943f1d5fc4/lib/ReactTags.js#L226 and https://github.com/prakhar1989/react-tags/blob/c59ff527f08b8f307f69d883be8bfe943f1d5fc4/lib/ReactTags.js#L230

jpbufe3:
The paste handler was simply splitting the text and passing it along, rather than creating the expected object type with an `id` and `text` field.

Fixes #305


jpbufe3:
@ad1992 - could you take a look at this PR as well? This is an actual bug that needs to be fixed.

test/reactTags.test.js
```diff
@@ -137,7 +137,7 @@ describe('Test ReactTags', () => {
       mockItem({
         delimiters: [Keys.TAB, Keys.SPACE, Keys.COMMA],
         handleAddition(tag) {
-          actual.push(tag);
+          actual.push(tag.text);
```
ad1992: Instead of changing it text here can you fix the spec so that it compares array of objects

lib/ReactTags.js
```diff
@@ -223,7 +223,9 @@ class ReactTags extends Component {
     const clipboardData = e.clipboardData || window.clipboardData;
     const string = clipboardData.getData('text');
     const regExp = new RegExp(`[${delimiterChars}]+`);
-    string.split(regExp).forEach((tag) => this.props.handleAddition(tag));
+    string
+      .split(regExp)
+      .forEach((tag) => this.props.handleAddition({ id: tag, text: tag }));
```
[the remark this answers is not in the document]
ad1992: `dasdf,df,adf,a,fa,df` pasting this will only render `df` as the tag in UI. can you fix this as well in this PR

test/reactTags.test.js
```diff
@@ -125,20 +125,21 @@ describe('Test ReactTags', () => {
     expect(actual).to.not.have.members(['Banana']);
   });
 
-  test('handles the paste event and splits the clipboard on delimiters', () => {
+  test('handles the paste event, splits the clipboard on delimiters, and ignores duplicates', () => {
```
ad1992: can you move the two tests related to paste event inside a new describe block 'tests handlePaste`

ad1992: rename this to `should split the clipboard on delimiters` and remove duplicate tags in this test. Add a new `it` block `should not allow duplicate tags` and test for duplicate tags there.

test/reactTags.test.js
```diff
@@ -148,19 +149,22 @@ describe('Test ReactTags', () => {
...
         getData: () =>
-          'Banana,Apple,Apricot\nOrange Blueberry,Pear,Peach\tKiwi',
+          'Banana,Apple,Apricot\nOrange Blueberry,Pear,Peach\tKiwi,Banana',
       },
     });
 
-    expect(actual).to.have.members([
-      'Banana',
+    // Note that 'Apple' and 'Banana' are only included once in the expected list
+    const expected = [
```
ad1992: remove `Apple` and `Banana` from this test and the comment as well

lib/ReactTags.js
```diff
@@ -223,7 +223,9 @@ class ReactTags extends Component {
     const clipboardData = e.clipboardData || window.clipboardData;
     const string = clipboardData.getData('text');
     const regExp = new RegExp(`[${delimiterChars}]+`);
-    string.split(regExp).forEach((tag) => this.props.handleAddition(tag));
+    string
```
ad1992: can you rename this `pastedText` ?

jpbufe3:
@ad1992 - this should be all set now. once this is merged, can we create `5.1.0` and release it?
thanks.

lib/ReactTags.js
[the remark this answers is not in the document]
ad1992: @jpbufe3 This issue is not fixed yet. Can you fix it ?

ad1992:
@jpbufe3 once this pr is approved and merged, I will surely release a new version.

lib/ReactTags.js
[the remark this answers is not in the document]
jpbufe3: yes it is. please look at the latest version of the code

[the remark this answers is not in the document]
jpbufe3: https://github.com/prakhar1989/react-tags/pull/306/files#diff-833c370f0b7e67628628a041d1984df3R228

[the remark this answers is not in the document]
ad1992: try passing the delimiters as `{[9, 32, 188]}` in example main.js file and paste `dasdf,df,adf,a,fa,df`. Only the last tag gets pasted.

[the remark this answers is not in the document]
jpbufe3: I'm sorry. I just tested again and it fails locally, but I'm not quite sure why given the test works. Will take another look.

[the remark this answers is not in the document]
jpbufe3: okay. so there's a bug in the `main.js` file for the `handleAddition` function. that being said, there's still a race condition with adding them all at once. I think they need to be de-duplicated in the paste method before sending them across either way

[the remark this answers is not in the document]
jpbufe3: this is now fixed. please note that anything that calls `setState` and uses a value from the state must use the callback form in order to avoid out of sync issues.

example/main.js
```diff
@@ -247,8 +248,7 @@ class App extends React.Component {
   }
 
   handleAddition(tag) {
-    const { tags } = this.state;
-    this.setState({ tags: [...tags, ...[tag]] });
+    this.setState(state => ({ tags: [...state.tags, tag] }));
```
ad1992: can you also pass the delimiters in this file to `Tags` component ?
you can make a constants.js file in the same directory and 
```js
 export const KEYCODES = {
   ENTER: 13,
   COMMA: 188,
}
```
and pass these delimiters.

lib/ReactTags.js
```diff
@@ -251,18 +251,15 @@ class ReactTags extends Component {
...
-      return;
-    }
-
-    e.preventDefault();
-
+  /**
+   * Used to determine how the pasted content is split.
+   * @returns {RegExp} Regular expression
+   */
+  get delimiterRegex() {
```
ad1992: any reason for using getter here ?
can we move this to a new file `utils.js` and rename this to `getDelimiterRegex(delimters)` ?

lib/ReactTags.js
```diff
@@ -251,18 +251,15 @@ class ReactTags extends Component {
...
     // See: http://stackoverflow.com/a/6969486/1463681
     const escapeRegex = (str) =>
```
ad1992: this can be made an independent function in utils.js and also doesn't need to be an arrow function

test/reactTags.test.js
```diff
@@ -96,66 +96,108 @@ describe('Test ReactTags', () => {
...
+      const $el = mount(
+        mockItem({
+          allowAdditionFromPaste: false,
+          handleAddition(tag) {
+            actual.push(tag);
+          },
+        })
+      );
+
+      const ReactTagsInstance = $el.instance().refs.child;
```
ad1992: this doesn't seem to be used anywhere so can be removed

test/reactTags.test.js
```diff
@@ -96,66 +96,108 @@ describe('Test ReactTags', () => {
...
+          },
+          tags,
+        })
+      );
 
-    $input.simulate('paste', {
-      clipboardData: {
-        getData: () => 'Banana',
-      },
+      const ReactTagsInstance = $el.instance().refs.child;
```
ad1992: this doesn't seem to be used anywhere so can be removed

test/reactTags.test.js
```diff
@@ -96,66 +96,108 @@ describe('Test ReactTags', () => {
...
+        })
+      );
 
-    const actual = [];
-    const $el = mount(
-      mockItem({
-        delimiters: [Keys.TAB, Keys.SPACE, Keys.COMMA],
-        handleAddition(tag) {
-          actual.push(tag);
+      const ReactTagsInstance = $el.instance().refs.child;
```
ad1992: this doesn't seem to be used anywhere so can be removed

lib/ReactTags.js
ad1992: @jpbufe3  once you create utils.js file can you also add specs for that file in tests folder ? You can write the specs in this PR or create a new PR as well for the specs once this pr is merged.

lib/ReactTags.js
jpbufe3: it prevents the regex from being re-generated every time someone goes to paste. happy to move it to a utils, but I think it should remain a getter in this class

ad1992: Hmm I get your point. But as its not an expensive operation which if regenerated will cause issue when it goes to paste and also this looks like something which can be reused  in future so IMO it can be moved to utils. Also if the value of props change in the parent component then the value has to be recalculated so I don't think we will get any benefits of using getter here as we are not memoizing it here. 
Also since you are not memoizing the result so the getter will be executed every time the handlePaste executes. Does this make sense  @jpbufe3 ?

ad1992:
@jpbufe3 can you make the pending changes ? Can't want to merge this :)

lib/ReactTags.js
ad1992: awesome. So you can directly import `buildRegexForDelimiters` from utils and directly use `buildRegexForDelimiters` in handlePaste.

jpbufe3:
I tried to create a separate constants file for the example, but was unable to import the values into the main file properly so I decided to just leave them at the top of the file for now. feel free to fix that separately if you'd like

lib/ReactTags.js
```diff
@@ -251,32 +252,27 @@ class ReactTags extends Component {
     }
   }
 
+  /**
+   * Used to determine how the pasted content is split.
+   * @returns {RegExp} Regular expression
+   */
+  get delimiterRegex() {
```
ad1992: this getter can be removed.

lib/utils.js
```diff
@@ -0,0 +1,18 @@
+export function escapeRegexString(regexString) {
+  // See: http://stackoverflow.com/a/6969486/1463681
+  return regexString.replace(/[-[\]/{}()*+?.\\^$|]/g, '\\$&');
+}
+
+/**
+ * Used to determine how the pasted content is split.
+ * @returns {RegExp} Regular expression
+ */
+export function convertDelimitersToRegex(delimiters) {
```
ad1992: can you rename this function to getDelimitersForRegex` or `buildDelimitersFromRegex` or `createDelimitersFromRegex`as its creating the regex with delimiters passed in params and not converting the delimiters to regex ?

lib/utils.js
```diff
@@ -0,0 +1,18 @@
+export function escapeRegexString(regexString) {
+  // See: http://stackoverflow.com/a/6969486/1463681
+  return regexString.replace(/[-[\]/{}()*+?.\\^$|]/g, '\\$&');
```
ad1992: I know you fixed the escape chars issue due to eslint here and thank you for that. But can you please revert the regex here. I want to push this regex change with good amount of testing and testcases so you can give another pr with only the regex change.

lib/ReactTags.js
```diff
@@ -251,32 +252,27 @@ class ReactTags extends Component {
...
-        })
-        .join('')
-    );
-
     const clipboardData = e.clipboardData || window.clipboardData;
-    const string = clipboardData.getData('text');
-    const regExp = new RegExp(`[${delimiterChars}]+`);
-    string.split(regExp).forEach((tag) => this.props.handleAddition(tag));
+    const pastedText = clipboardData.getData('text');
+    const tags = pastedText.split(this.delimiterRegex);
```
ad1992: directly use the function here instead of the getter

ad1992:
@jpbufe3 awesome work in fixing typos in eslint config and upgrading eslint in codeclimate 👍

lib/utils.js
jpbufe3: I found a better solution: https://lodash.com/docs#escapeRegExp
no need for testing this way

lib/utils.js
jpbufe3: I'm not sure I understand. You are taking the `delimiters` and turning (converting) them into a regex.

all of your names make it sound to me like you are creating the delimiters rather than the regex, whereas my name is much more clear as to what the input and output is

ad1992: I m sorry my bad. I made a typo here. I meant can you rename this function to `getRegexForDelimiters` or `buildRegexFromDelimiters` or `createRegexFromDelimiters` as you said its creating the regex from the delimiters.

lib/utils.js
```diff
@@ -0,0 +1,17 @@
...
+ * @param {Array<char>} delimiters Array of characters to turn into a regex
+ * @returns {RegExp} Regular expression
+ */
+export function convertDelimitersToRegExp(delimiters) {
+  const delimiterChars = delimiters.map((delimiter) => {
+    // See: http://stackoverflow.com/a/34711175/1463681
+    const chrCode = delimiter - 48 * Math.floor(delimiter / 48);
+    return String.fromCharCode(96 <= delimiter ? chrCode : delimiter);
+  }).join('');
+  const escapedDelimiterChars = escapeRegExp(delimiterChars);
```
ad1992: but the lodash doesn't escape `/` which is bad as it will lead to invalid regex
when string is `'[lodash](https://lodash.com/)'`
output of lodash
`"\[lodash\]\(https://lodash\.com/\)"`
output of regex we use is
`\[lodash\]\(https:\/\/lodash\.com\/\)`

jpbufe3: I just added a test to prove that is not true.

ad1992:
@jpbufe3 since you made multiple changes in this single PR can you actually `squash` or `fixup` the commits and keep only 3 commits. you can reword the commit messages the way you want, below is just an example

* Fix pasting in react tag input
* upgrade eslint and fix typos
* Added buildRegExpFromDelimiters in util with specs

jpbufe3:
I made it 4 commits to add in the example update. Let me know if you'd like me to merge the last two commits though.

PR #308, opened 2018-04-27
Move the props and default props into the class

jpbufe3:
Per request from @ad1992 on #304, moving the prop types and default props into the `ReactTags` class and making them static members.


ad1992:
@jpbufe3 thanks for the contribution.

PR #311, opened 2018-05-04
Changes suggestions to use the provided labelField

josh-degraw:
This pull request changes Suggestions to use the provided `labelField` instead of always using hardcoded `.text` property, so that the suggestions and tags don't have to be mapped differently.


ad1992:
@josh-degraw  seems like the tests are failing. Can you fix the tests and push again

josh-degraw:
Sorry about that! Fixed all of my issues, and to make it pass I had to fix some things that were already in there, most of which were eslint warnings. These had absolutely nothing to do with what I've added, but were preventing the CI from passing. They were all very minor things (sorting of functions / proptypes, etc.), so i fixed some and disabled some inline. No functionality was changed with these, but I just wanted to make sure it's all kosher and whatnot.

ad1992:
@josh-degraw I wanted you to fix only the travis tests. I recently added the pre-commit and integrated the eslint with code climate due to which you would have got the code climate issues Sorry for that. I am already working on fixing the code climate issues but since you have fixed lot of it in this pr, Can you please move all those changes which are not relevant to your PR in a different PR ? Just make sure the travis test pass in this PR

josh-degraw:
Okay, I think I removed all the stuff not directly related to the issue at hand here. I'll submit another PR with the lint stuff.

ad1992:
@josh-degraw can you remove package-lock.json from the PR ?

josh-degraw:
@ad1992 Done.

lib/ReactTags.js
```diff
@@ -75,6 +78,7 @@ class ReactTags extends Component {
       selectionMode: false,
     };
 
+    this.addTag = this.addTag.bind(this);
```
ad1992: why binding here ?

lib/ReactTags.js
```diff
@@ -84,6 +88,7 @@ class ReactTags extends Component {
     this.resetAndFocusInput = this.resetAndFocusInput.bind(this);
     this.handleSuggestionHover = this.handleSuggestionHover.bind(this);
     this.handleSuggestionClick = this.handleSuggestionClick.bind(this);
+    this.filteredSuggestions = this.filteredSuggestions.bind(this);
```
ad1992: why binding here ?

lib/ReactTags.js
```diff
@@ -8,16 +8,17 @@ import Tag from './Tag';
 import { buildRegExpFromDelimiters } from './utils';
 
 //Constants
-import { KEYS, DEFAULT_PLACEHOLDER, DEFAULT_CLASSNAMES } from './constants';
+import { KEYS, DEFAULT_PLACEHOLDER, DEFAULT_CLASSNAMES, DEFAULT_LABEL_FIELD } from './constants';
```
ad1992: please run `npm run format`

lib/Suggestions.js
```diff
@@ -67,8 +68,9 @@ class Suggestions extends Component {
 
   markIt = (input, query) => {
     const escapedRegex = query.trim().replace(/[-\\^$*+?.()|[\]{}]/g, '\\$&');
+    const html = input[this.props.labelField];
```
ad1992: `const { [this.props.labelField]: labelField } = input;`

lib/Suggestions.js
```diff
@@ -67,8 +68,9 @@ class Suggestions extends Component {
...
     return {
-      __html: input.text.replace(RegExp(escapedRegex, 'gi'), '<mark>$&</mark>'),
+      __html: html.replace(RegExp(escapedRegex, 'gi'), '<mark>$&</mark>'),
```
ad1992: labelField.replace

lib/ReactTags.js
josh-degraw: To prevent issues when calling this.props

lib/ReactTags.js
josh-degraw: To prevent issues when calling this.props

lib/ReactTags.js
ad1992: you can use arrow functions in that case :)

lib/ReactTags.js
ad1992: same as above arrow functions

ad1992:
@josh-degraw any updates ?

lib/Suggestions.js
josh-degraw: I'm confused as to what you're wanting here and where you put
> ``` js 
> const { [this.props.labelField]: labelField } = input;
> ```

The labelField isn't what's being put into the `__html` field. Could you clarify for me?

josh-degraw: Yeah I get what it's doing, it was really the naming you put that was confusing. `labelFieldValue` is still a bit ambiguous to me..
What about 
```js
const { [this.props.labelField]: text } = input;
```
or 
```js
const { [this.props.labelField]: labelValue} = input;
```

It's really just semantics, is should just be clear to anyone what the actual value is.

ad1992: Yeah sorry my bad. `const { [this.props.labelField]: labelValue} = input;` this looks fine 👍

test/reactTags.test.js
```diff
@@ -12,6 +12,7 @@ import { KEYS, DEFAULT_PLACEHOLDER } from '../lib/constants';
 
 const defaults = {
   tags: [{ id: 'Apple', text: 'Apple' }],
+  labelField: 'text',
```
ad1992: this should not be needed as its mentioned in default prop

lib/Suggestions.js
ad1992: this change seems to be pending

josh-degraw:
@ad1992 Okay I think it should be good to go. What exactly do you mean by adding specs?

package-lock.json
```diff
@@ -59,10 +58,7 @@
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/@sinonjs/formatio/-/formatio-2.0.0.tgz",
       "integrity": "sha512-ls6CAMA6/5gG+O/IdsBcblvnd8qcO/l1TYoNeAzp3wcISOxlPXQEus0mLcdwazEkWjaBdaJ3TaxmNgCLWwvWzg==",
-      "dev": true,
-      "requires": {
-        "samsam": "1.3.0"
-      }
+      "dev": true
```
ad1992: I don't see any change in package.json. this file shouldn't be committed. Can you remove this

lib/ReactTags.js
```diff
@@ -107,15 +115,19 @@ class ReactTags extends Component {
     }
   }
 
-  filteredSuggestions(query, suggestions) {
+  filteredSuggestions = (query, suggestions) => {
```
ad1992: We don't need arrow function here. Can this be removed ?

ad1992:
@josh-degraw I meant can you add specs with custom label field since thats the change in this PR ? Currently, only the default label field is added in the specs but there is no spec with custom label fields.
You can add specs with custon label field and test the changes made in the code.

lib/ReactTags.js
josh-degraw: Probably. The only reason I did it like this is because you said you didn't want to bind it and to use arrow functions instead.

josh-degraw:
@ad1992 So are you talking about adding documentation for using a custom label field in the readme?

lib/ReactTags.js
ad1992: yeah but binding was also not needed here thats why adding arrow function was not helping. It was needed here https://github.com/josh-degraw/react-tags/blob/1a9ad5470ceb98a0584fb5f80964146b0b4cb350/lib/ReactTags.js#L123 which is fine.

package-lock.json
```diff
@@ -12919,4 +12919,4 @@
       }
     }
   }
-}
+}
```
ad1992: seems like you removed new line thats why this file is still showing in diff. can you fix this ?

ad1992:
@josh-degraw yes the documentation in read me is needed. But what I meant by adding specs is adding the necessary tests in `reactTags.test.js` and `suggestions.test.js` files with custom label field. Currently you are only passing `labelField` as `DEFAULT_LABEL_FIELD` https://github.com/josh-degraw/react-tags/blob/27e45d9565973cb486f90cf3a770ccffa05e5ba0/test/reactTags.test.js#L44

So specs like if suggestions are filtering properly since you made small change there, then if tags are getting added properly, only uniq tags are allowed etc when the the prop is passed. Basically its better to add specs for the changes we make so that we can be sure that its not breaking anything.

In case you still have any confusion, let me know.

ad1992:
@josh-degraw sure I will check the tests. Can you please rebase with master and fix the conflicts?

test/reactTags.test.js
```diff
@@ -494,4 +495,23 @@ describe('Test ReactTags', () => {
       expect(actual).to.have.deep.members([{ id: 'Ea', text: 'Ea' }]);
     });
   });
+
+  test('Utilizes custom labelField properly', () => {
+    const testLabelField = 'name';
+    const mapper = (a) => ({ id: a.id, [testLabelField]: a.text });
```
ad1992: `(data) => ({ id: data.id, name: data.text });`

test/reactTags.test.js
```diff
@@ -494,4 +495,23 @@ describe('Test ReactTags', () => {
...
+    const suggestions = defaults.suggestions.map(mapper);
+    const tags = defaults.tags.map(mapper);
+    const expectedText = tags[0][testLabelField];
+
+    const props = {
+      labelField: testLabelField,
```
ad1992: `labelField`,

test/reactTags.test.js
```diff
@@ -494,4 +495,23 @@ describe('Test ReactTags', () => {
       expect(actual).to.have.deep.members([{ id: 'Ea', text: 'Ea' }]);
     });
   });
+
+  test('Utilizes custom labelField properly', () => {
```
ad1992: `should add tag with custom label field`

test/reactTags.test.js
```diff
@@ -494,4 +495,23 @@ describe('Test ReactTags', () => {
...
+    const expectedText = tags[0][testLabelField];
+
+    const props = {
+      labelField: testLabelField,
+      suggestions,
+      tags,
+    };
+
+    const $el = mount(mockItem(props));
+    expect($el.prop('labelField')).to.equal(testLabelField);
```
ad1992: this can be removed as we are passing the same prop and checking it.
instead lets simulate `keyDown` to check if correct tag is added.
You can check this for writing the test https://github.com/josh-degraw/react-tags/blob/175887f22d908b2a7a9514ae88dc32571249506f/test/reactTags.test.js#L242

josh-degraw:
Okay I think I made all the changes you asked. Let me know if anything else is needed.

README.md
```diff
@@ -158,37 +158,42 @@ Option | Type | Default | Description
...
 const tags =  [ { id: "1", text: "Apples" } ]
+
+// With labelField of `name`
+const tags =  [ { id: "1", name: "Apples" } ]
 ```
 
 <a name="suggestionsOption"></a>
 ##### suggestions (optional, defaults to `[]`)
-An array of suggestions that are used as basis for showing suggestions. At the moment, this should be an array of strings.
+An array of suggestions that are used as basis for showing suggestions. These objects should follow the same structure as the data. So if the `labelField` is `name`, the following would work:
```
ad1992: These objects should follow the same structure as the  [`tags`](#tagsOption).
So if the `labelField` is `name`, the following would work:

josh-degraw:
@ad1992 Okay, I reverted the changes from `package-lock.json` and fixed the `README.md` issues and pushed up. I also double-checked and you should have edit permissions on this PR. Hope it's good to go! Thanks for working with me on this! This has been my first _real_ PR so I appreciate your patience and help with this.

test/suggestions.test.js
```diff
@@ -5,6 +5,8 @@ import { shallow, mount } from 'enzyme';
 import { spy } from 'sinon';
 import Suggestions from '../lib/Suggestions';
 import noop from 'lodash/noop';
+import { DEFAULT_LABEL_FIELD } from '../lib/constants';
+import {} from './utils.test';
```
ad1992: this can be removed.

ad1992:
@josh-degraw seems like there is some permission issue. can you push https://github.com/prakhar1989/react-tags/compare/josh-degraw-master?expand=1 to your master and reopen this PR ?

josh-degraw:
Okay, I cherry-picked that commit and pushed it to my `master`. I'm not sure how to reopen the PR though. Do you mean create a new one?

ad1992:
@josh-degraw can you squash only all your commits into one commit and push to this branch again.

josh-degraw:
Thank you!

PR #339, opened 2018-06-05
Import only wanted parts of lodash

tabrath:
* By being explicit about what functions we want from lodash, we can reduce the bundle size by ~50%.
* Adding 'sideEffects: false' to package.json to allow webpack4 optimize re-exports (https://github.com/webpack/webpack/tree/master/examples/side-effects).

Before:
* ReactTags.min.js 143 kB

After:
* ReactTags.min.js 72.4 kB

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "5.0.2",
   "description": "React tags is a fantastically simple tagging component for your React projects",
   "main": "dist-modules/ReactTags.js",
+  "sideEffects": false,
   "scripts": {
     "test": "jest --notify --coverage",
     "test-watch": "jest --watch --notify --coverage",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "5.0.2",
   "description": "React tags is a fantastically simple tagging component for your React projects",
   "main": "dist-modules/ReactTags.js",
+  "sideEffects": false,
```
ad1992: @tabrath currently we are at webpack `3.11.0` so I guess this flag won't be working .  Can you remove this from this pr and give a PR with updating webpack and also adding this flag ?

tabrath:
Ok, I’ll do that.

ad1992:
Thanks for your contribution @tabrath

PR #344, opened 2018-06-07
fix autocomplete bug

### LINKED ISSUE #342: autocomplete is bugged?
Hi, I was trying to use autocomplete in my application and it didn't work, it produced an error in the console instead. I couldn't find an issue with my code so I decided to try implementing autocomplete in your example code. The same error occured:

Uncaught TypeError: query.toLowerCase is not a function
    at ReactTags.js:116
    ...

Please have a look as I would like to have the autocomplete for my application. Thanks and cheers!

tabrath:
The `tag` object was sent to `filteredSuggestions`, but the actual text was expected. Sending `tag.text` fixes the issue. Closes #342

### DIFF
--- a/lib/ReactTags.js
+++ b/lib/ReactTags.js
@@ -282,7 +282,7 @@ class ReactTags extends Component {
     }
     if (this.props.autocomplete) {
       const possibleMatches = this.filteredSuggestions(
-        tag,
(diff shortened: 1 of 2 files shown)


ad1992:
Thanks @tabrath

PR #380, opened 2018-07-30
Change propTypes for maxLength

budaaaa:
While implementing the library i noticed that maxLength requires a number as a prop but in ReactTags.js a string is required.

### DIFF
--- a/lib/ReactTags.js
+++ b/lib/ReactTags.js
@@ -47,7 +47,7 @@ class ReactTags extends Component {
     classNames: PropTypes.object,
     name: PropTypes.string,
     id: PropTypes.string,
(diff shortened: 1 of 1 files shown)


ad1992:
@budaaaa the prop will work when its string since we are just passing the prop to input element. Can you instead change the proptype in  docs ?

budaaaa:
Hello @ad1992 
Tested it and it works either way.

Isn't it better if we keep it as a type number, cause like someone can add a letter by mistake and it wouldn't throw a warning.

ad1992:
yes sounds good. Thanks for your contribution @budaaaa

PR #413, opened 2018-10-09
Allowed characters

przpl:
_allowedCharacters_ property defines which characters are allowed in input.

### DIFF
--- a/src/ReactTags.js
+++ b/src/ReactTags.js
@@ -55,6 +55,7 @@ class ReactTags extends Component {
       })
     ),
     allowUnique: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


src/ReactTags.js
```diff
@@ -286,6 +288,15 @@ class ReactTags extends Component {
     );
   }
 
+  handleKeyPress(e) {
```
ad1992: @przpl  thanks for the contribution. We already have a `handleInputChange` callback which gets triggered when input changes so this validation can be handled at parent component. IMO this validation should handled at parent level. What do you think ?

PR #414, opened 2018-10-19
Escape suggestions before including them in HTML

devinlundberg:
Fixes potential XSS vulnerabilities as described in: https://github.com/prakhar1989/react-tags/issues/223

### DIFF
--- a/__tests__/reactTags.test.js
+++ b/__tests__/reactTags.test.js
@@ -584,8 +584,10 @@ test('should allow duplicate tags when allowUnique is false', () => {
   const $input = $el.find('.ReactTags__tagInputField');
   $input.simulate('change', { target: { value: 'Apple' } });
   $input.simulate('keyDown', { keyCode: ENTER_ARROW_KEY_CODE });
-  expect(actual).to.have.deep.members([{
-    id: 'Apple',
-    text: 'Apple',
-  }]);
+  expect(actual).to.have.deep.members([
+    {
+      id: 'Apple',
+      text: 'Apple',
+    },
+  ]);
 });
--- a/__tests__/suggestions.test.js
+++ b/__tests__/suggestions.test.js
@@ -78,7 +78,7 @@ describe('Suggestions', function() {
     expect($el.find('li').length).to.equal(0);
   });
 
-  test('should not render html input in query or suggestions', function() {
+  test('should escape html characters in query', function() {
     const suggestions = [{ id: 'script', text: '<script>alert()</script>' }];
     const $el = shallow(
       mockItem({
@@ -86,7 +86,6 @@ describe('Suggestions', function() {
         suggestions,
       })
     );
-    $el.setProps({ suggestions });
     expect($el.html()).to.equal(
       '<div class="foo"><ul> <li class=""><span><mark>&lt;script&gt;alert()&lt;/script&gt;</mark></span></li> </ul></div>'
     );
--- a/__tests__/suggestions.test.js
+++ b/__tests__/suggestions.test.js
@@ -78,6 +78,20 @@ describe('Suggestions', function() {
     expect($el.find('li').length).to.equal(0);
   });
 
+  test('should not render html input in query or suggestions', function() {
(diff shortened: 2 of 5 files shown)


src/Suggestions.js
```diff
@@ -75,7 +75,15 @@ class Suggestions extends Component {
     const { [this.props.labelField]: labelValue } = input;
 
     return {
-      __html: labelValue.replace(RegExp(escapedRegex, 'gi'), '<mark>$&</mark>'),
+      __html: labelValue.replace(RegExp(escapedRegex, 'gi'), (x) => {
+        const escapedInput = x
+          .replace(/&/g, '&amp;')
```
ad1992: @devinlundberg can you move the escape logic to the `utils.js` and add spec for util as well ?

src/Suggestions.js
```diff
@@ -75,7 +76,9 @@ class Suggestions extends Component {
     const { [this.props.labelField]: labelValue } = input;
 
     return {
-      __html: labelValue.replace(RegExp(escapedRegex, 'gi'), '<mark>$&</mark>'),
+      __html: labelValue.replace(RegExp(escapedRegex, 'gi'), (x) => {
+        return '<mark>' + escapeHTML(x) + '</mark>';
```
ad1992: template string ?

test/suggestions.test.js
```diff
@@ -78,6 +78,20 @@ describe('Suggestions', function() {
...
 
+  test('should not render html input in query or suggestions', function() {
+    const suggestions = [{ id: 'script', text: '<script>alert()</script>' }];
+    const $el = shallow(
+      mockItem({
+        query: '<script>alert()</script>',
+        suggestions: suggestions,
+      })
+    );
+    $el.setProps({ suggestions: suggestions });
```
ad1992: `$el.setProps({ suggestions });`

test/suggestions.test.js
```diff
@@ -78,6 +78,20 @@ describe('Suggestions', function() {
     expect($el.find('li').length).to.equal(0);
   });
 
+  test('should not render html input in query or suggestions', function() {
+    const suggestions = [{ id: 'script', text: '<script>alert()</script>' }];
+    const $el = shallow(
+      mockItem({
+        query: '<script>alert()</script>',
+        suggestions: suggestions,
```
ad1992: `suggestions`,

src/utils.js
```diff
@@ -17,3 +17,17 @@ export function buildRegExpFromDelimiters(delimiters) {
   const escapedDelimiterChars = escapeRegExp(delimiterChars);
   return new RegExp(`[${escapedDelimiterChars}]+`);
 }
+
+/**
+ * Escape text to be used in HTML.
+ * @param {string} text to be escaped
+ * @returns {string} HTML
+ */
+export function escapeHTML(text) {
```
ad1992: `lodash` is already added in this library so can we use `escape` in lodash so that the custom escape can be omitted ? in that case it can be directly used `Suggestion.js`

test/suggestions.test.js
```diff
@@ -78,6 +78,20 @@ describe('Suggestions', function() {
...
 
+  test('should not render html input in query or suggestions', function() {
+    const suggestions = [{ id: 'script', text: '<script>alert()</script>' }];
+    const $el = shallow(
+      mockItem({
+        query: '<script>alert()</script>',
+        suggestions,
+      })
+    );
+    $el.setProps({ suggestions });
```
ad1992: this should not be required as you are setting suggestions above

test/suggestions.test.js
```diff
@@ -78,6 +78,20 @@ describe('Suggestions', function() {
     expect($el.find('li').length).to.equal(0);
   });
 
+  test('should not render html input in query or suggestions', function() {
```
ad1992: `should escape html characters in query`

devinlundberg:
@ad1992  can you please take another look? I believe all of you requested changes were resolved.

ad1992:
@devinlundberg thanks for your contribution !!

PR #420, opened 2018-10-28
Enable an option to have empty tags

sp2hari:


__tests__/reactTags.test.js
```diff
@@ -589,3 +590,38 @@ test('should allow duplicate tags when allowUnique is false', () => {
...
+        },
+        allowEmpty: allowEmpty,
+      })
+    );
+
+    expect($el.instance().props.tags).to.have.deep.members(defaults.tags);
+    const $input = $el.find('.ReactTags__tagInputField');
+    $input.simulate('change', { target: { value: '' } });
+    $input.simulate('keyDown', { keyCode: ENTER_ARROW_KEY_CODE });
+    if (allowEmpty) {
```
ad1992: noooo lets not have if else conditions for assertions inside specs. This approach is fine when the assertions are exactly same. Please create two different specs

src/components/ReactTags.js
```diff
@@ -234,7 +236,7 @@ class ReactTags extends Component {
           ? suggestions[selectedIndex]
           : { id: query, [this.props.labelField]: query };
 
-      if (selectedQuery !== '') {
+      if (selectedQuery !== '' || this.props.allowEmpty === true) {
```
ad1992: why ` this.props.allowEmpty === true`, its anyways returning when `allowEmpty` is false

src/components/ReactTags.js
```diff
@@ -292,8 +294,8 @@ class ReactTags extends Component {
   }
 
   addTag = (tag) => {
-    const { tags, labelField, allowUnique } = this.props;
-    if (!tag.id || !tag[labelField]) {
+    const { tags, labelField, allowUnique, allowEmpty } = this.props;
+    if ( (!tag.id || !tag[labelField]) && allowEmpty === false) {
```
ad1992: instead of doing `&&` here can we do early return when `!allowEmpty` ?

PR #456, opened 2018-12-26
Update class

harshithmullapudi:
Now when the className props get updated it gets applied for the components also

### DIFF
--- a/src/components/ReactTags.js
+++ b/src/components/ReactTags.js
@@ -29,7 +29,7 @@ class ReactTags extends Component {
       PropTypes.shape({
         id: PropTypes.string.isRequired,
       })
-    ),
+      ),
     delimiters: PropTypes.arrayOf(PropTypes.number),
     autofocus: PropTypes.bool,
     inline: PropTypes.bool,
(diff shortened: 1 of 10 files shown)


ad1992:
@harshithmullapudi this pr seems to have changes which are already made in https://github.com/prakhar1989/react-tags/pull/453
So either raise the pr against `harshithmullapudi:master` or cherry-pick the commits needed and  push to the previous pr.

ad1992:
@harshithmullapudi can you add specs for this

src/components/ReactTags.js
```diff
@@ -105,6 +105,15 @@ class ReactTags extends Component {
     this.handleSuggestionClick = this.handleSuggestionClick.bind(this);
   }
 
+  static getDerivedStateFromProps(props)
+  {
+    const { classNames } = props;
+
+    return {
+      classNames : {...DEFAULT_CLASSNAMES,...classNames},
```
ad1992: move `...DEFAULT_CLASSNAMES`  to constructor.
use [memoize-one](https://www.google.com/search?q=memoize_one&oq=memo&aqs=chrome.0.69i59j69i57j69i60j0l3.1328j1j7&sourceid=chrome&ie=UTF-8) here so that state changes only when className prop changes.

src/components/ReactTags.js
```diff
@@ -21,6 +21,14 @@ import {
 
 import '../styles/react-tags.scss';
 
+const updateClass  = (classNames) =>
```
ad1992: can you add a TODO in line 104 to remove classNames from state and change updateClass to instance function.

ad1992:
@harshithmullapudi thanks for your contribution

PR #474, opened 2019-01-27
Add event at handleAddition

melenaos:


ad1992:
@melenaos can you run `npm run format` to fix the formatting issues

melenaos:
I have run it but there where no changes. Is it something specific that has formatting problem?

src/components/ReactTags.js
```diff
@@ -293,11 +293,11 @@ class ReactTags extends Component {
 
     // Only add unique tags
     uniq(tags).forEach((tag) =>
-      this.addTag({ id: tag, [this.props.labelField]: tag })
+      this.addTag({ id: tag, [this.props.labelField]: tag }, e)
     );
   }
 
-  addTag = (tag) => {
+    addTag = (tag, e) => {
```
ad1992: remove leading space

src/components/ReactTags.js
```diff
@@ -323,7 +323,7 @@ class ReactTags extends Component {
     }
 
     // call method to add
-    this.props.handleAddition(tag);
+    this.props.handleAddition(tag, e);
```
ad1992: can you also add a spec for this change ?

ad1992:
@melenaos any update?

src/components/ReactTags.js
melenaos: What spec to add? More comment than what it already has?

melenaos:
> 
> 
> @melenaos any update?

I will commit it today.
I have made the space change and I am waiting for your reply on the spec change

melenaos:
I commit one change, is it automatically added to the merge request?

src/components/ReactTags.js
ad1992: @melenaos spec to check if the prop `handleAddition` was called with `tag` and `event`. To give you an idea how it will look, you can check here.
You can stub the method `handleAddition` and simulate `keydown` event which will trigger the `handleAddition` prop to be called.
you can check https://github.com/prakhar1989/react-tags/blob/17d912747a600295242209ae5c375f85dc790a52/__tests__/reactTags.test.js#L529
it will be similar to this spec but instead if passing the handleAddition function, just stub it and check if it was called with the expected params.

ad1992: @melenaos any update ?

Issue #486, opened 2019-02-06
`autocomplete` and `addTag` logic broken by default


ZebGirouard:
At the moment, `ReactTags.js`, on line 466 has this conditional:

```
    if (_this5.props.autocomplete) {
      var possibleMatches = _this5.filteredSuggestions(tag[labelField], _this5.props.suggestions);

      if (_this5.props.autocomplete === 1 && possibleMatches.length === 1 || _this5.props.autocomplete === true && possibleMatches.length) {
        tag = possibleMatches[0];
      }
    }
```

If you look closely, you'll notice a typo...at least I think it's a typo, because it causes tag suggestion selection to ALWAYS select the first element in a list of suggestions.

`&& possibleMatches.length` <== that should (I'm like 99% sure) say `&& possibleMatches.length === 1`

My workaround for now is to set the prop `autocomplete={ 1 }`, because `autocomplete` causes the first item in a list to be selected, whether I click on the second, third, fourth, etc item.

Do y'all want me to file a PR for this?  It's 6 characters including spaces, and would probably help a lot of people.

ZebGirouard:
Hope that PR is all you need...full project search did not show me `possibleMatches.length` anywhere else, but with no feedback, I had to guess that "my fix" is "the fix".

ad1992:
@ZebGirouard sorry for the late reply and thanks for the PR. I agree with you that first element should only be selected when only one match is found but I could not understand 
`autocomplete causes the first item in a list to be selected, whether I click on the second, third, fourth, etc item.`
Do you mean when you pass `autocomplete` as true and while typing if there are multiple matches and you select some `x` element but in UI it shows up first element ?

ZebGirouard:
What a bummer.  I really hope this issue was fixed somewhere else.  To be completely frank, I am a little disappointed that this issue remains unfixed because I was unsure how to add a test case to a project I've never touched before.  I think it cannot be stressed enough how many people fall off open source projects before they get a chance to really help. Not sure why someone else couldn't have pushed up a `spec` change on my branch, or given me some advice on how to complete it, in the two months this issue and PR have been open.

PR #492, opened 2019-02-15
Fixes Issue #486 - length should be equal to 1 to pick first tag

ZebGirouard:
Fixes what I assume to be a typo.  Whether the value of `autocomplete` is equal to `1` or `true`, we should only default to the first element if there is only one element.  `1` vs `true` should not change this behavior.

fixes https://github.com/prakhar1989/react-tags/issues/486

### DIFF
--- a/src/components/ReactTags.js
+++ b/src/components/ReactTags.js
@@ -316,7 +316,7 @@ class ReactTags extends Component {
 
       if (
         (this.props.autocomplete === 1 && possibleMatches.length === 1) ||
-        (this.props.autocomplete === true && possibleMatches.length)
+        (this.props.autocomplete === true && possibleMatches.length === 1)
       ) {
         tag = possibleMatches[0];
       }
--- a/src/components/ReactTags.js
+++ b/src/components/ReactTags.js
@@ -315,8 +315,7 @@ class ReactTags extends Component {
       );
 
       if (
-        (this.props.autocomplete === 1 && possibleMatches.length === 1) ||
-        (this.props.autocomplete === true && possibleMatches.length === 1)
+        this.props.autocomplete && possibleMatches.length === 1
       ) {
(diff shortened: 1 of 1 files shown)


src/components/ReactTags.js
```diff
@@ -316,7 +316,7 @@ class ReactTags extends Component {
 
       if (
         (this.props.autocomplete === 1 && possibleMatches.length === 1) ||
-        (this.props.autocomplete === true && possibleMatches.length)
+        (this.props.autocomplete === true && possibleMatches.length === 1)
```
ad1992: can you remove this line and instead change line 318 to 
```
this.props.autocomplete && possibleMatches.length === 1
```
This will take care of both the cases.
Please add a spec for this change.

ad1992:
@ZebGirouard any update ?

ZebGirouard:
@ad1992 , all set.  Thanks for the rec!

src/components/ReactTags.js
```diff
@@ -315,8 +315,7 @@ class ReactTags extends Component {
       );
 
       if (
-        (this.props.autocomplete === 1 && possibleMatches.length === 1) ||
-        (this.props.autocomplete === true && possibleMatches.length)
+        this.props.autocomplete && possibleMatches.length === 1
```
ad1992: @ZebGirouard can you add spec for this as well ?

ad1992: @ZebGirouard any update ?

ZebGirouard: Hi @ad1992 , I changed line 318 as requested, am I missing something?

ZebGirouard: Oh, are you saying I need to add a test?  Not sure how to do that as I'm new to this project....

Issue #352, opened 2018-06-22
Partial word search


gaurav3017:
Is it possible to partial search a tag? For example, if I type "Lanka", it should suggest "Sri Lanka". Right now, "Sri Lanka" is only suggested when "Sri" is typed.

siliconeidolon:
You'll need to write your own function for `handleFilterSuggestions`, eg: 
```// For allowing searching within strings
	handleFilterSuggestions(textInputValue, suggestions) {
		const lowerCaseQuery = textInputValue.toLowerCase();
		const filteredSuggestions = suggestions.filter(suggestion => {
			return suggestion.toLowerCase().includes(lowerCaseQuery);
		})

		return filteredSuggestions.sort((a, b) => {
			a = a.toLowerCase();
			b = b.toLowerCase();

			return a < b ? -1 : a > b ? 1 : 0;
		});
	}```

ad1992:
@gaurav3017 we don't support it right now but would be happy to receive a PR for this.

PR #495, opened 2019-02-16
Add partial word search

Gibr:
Add Partial word search according with issue #352 

Note: at all suggestions list first displayed suggestions with first match characters, after that displayed suggestions with partial match

### DIFF
--- a/__tests__/reactTags.test.js
+++ b/__tests__/reactTags.test.js
@@ -426,7 +426,7 @@ describe('Test ReactTags', () => {
         defaults.suggestions
       );
 
-      $input.simulate('change', { target: { value: 'ea' } });
+      $input.simulate('change', { target: { value: 'or' } });
       expect(ReactTagsInstance.state.suggestions).to.have.members([]);
 
       $input.simulate('change', { target: { value: 'ap' } });
@@ -538,10 +538,10 @@ describe('Test ReactTags', () => {
       );
       const $input = $el.find('.ReactTags__tagInputField');
 
-      $input.simulate('change', { target: { value: 'Ea' } });
+      $input.simulate('change', { target: { value: 'Or' } });
       $input.simulate('focus');
       $input.simulate('keyDown', { keyCode: ENTER_ARROW_KEY_CODE });
-      expect(actual).to.have.deep.members([{ id: 'Ea', text: 'Ea' }]);
+      expect(actual).to.have.deep.members([{ id: 'Or', text: 'Or' }]);
     });
     test('should add tag with custom label field and default suggestion filter', () => {
       const labelField = 'name';
@@ -562,10 +562,10 @@ describe('Test ReactTags', () => {
         })
       );
       const $input = $el.find('.ReactTags__tagInputField');
-      $input.simulate('change', { target: { value: 'Ea' } });
+      $input.simulate('change', { target: { value: 'Or' } });
       $input.simulate('focus');
       $input.simulate('keyDown', { keyCode: ENTER_ARROW_KEY_CODE });
-      expect(actual).to.have.deep.members([{ id: 'Ea', name: 'Ea' }]);
+      expect(actual).to.have.deep.members([{ id: 'Or', name: 'Or' }]);
     });
     test('should select the correct suggestion using the keyboard when label is custom', () => {
       const labelField = 'name';
--- a/__tests__/reactTags.test.js
+++ b/__tests__/reactTags.test.js
@@ -429,6 +429,12 @@ describe('Test ReactTags', () => {
       $input.simulate('change', { target: { value: 'or' } });
       expect(ReactTagsInstance.state.suggestions).to.have.members([]);
 
+      $input.simulate('change', { target: { value: 'ea' } });
+      expect(ReactTagsInstance.state.suggestions).to.have.deep.members([
+        { id: 'Pear', text: 'Pear' },
+        { id: 'Peach', text: 'Peach' },
+      ]);
+
       $input.simulate('change', { target: { value: 'ap' } });
       expect(ReactTagsInstance.state.suggestions).to.have.deep.members([
         { id: 'Apple', text: 'Apple' },
--- a/src/components/ReactTags.js
+++ b/src/components/ReactTags.js
@@ -131,19 +131,15 @@ class ReactTags extends Component {
       return this.props.handleFilterSuggestions(query, suggestions);
     }
 
+    const queryIndex = (item) =>
+      item[this.props.labelField]
+        .toLowerCase()
+        .indexOf(query.toLowerCase());
     return suggestions.filter((item) => {
-      return (
-        item[this.props.labelField]
-          .toLowerCase()
-          .indexOf(query.toLowerCase()) === 0
-      );
+      return queryIndex(item) === 0;
     })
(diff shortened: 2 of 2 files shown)


Gibr:
Failed some test for empty search result - now it finds partial suggestions.
It seems i have to rewrite test cases, isn't it?

ad1992:
thanks for the PR @Gibr, yes you need to fix the specs. It would be better to fix the code climate issue unless its not fixable.

ad1992:
ping @Gibr

Gibr:
Test cases were edited, code climate errors were fixed.

src/components/ReactTags.js
```diff
@@ -131,13 +131,16 @@ class ReactTags extends Component {
...
-      return (
-        item[this.props.labelField]
-          .toLowerCase()
-          .indexOf(query.toLowerCase()) === 0
-      );
-    });
+      return queryIndex(item) === 0;
+    })
+      .concat(suggestions.filter((item) => {
+        return queryIndex(item) > 0;
```
ad1992: both of this can be merged to `queryIndex(item) >= 0`

__tests__/reactTags.test.js
```diff
@@ -538,10 +538,10 @@ describe('Test ReactTags', () => {
       );
       const $input = $el.find('.ReactTags__tagInputField');
 
-      $input.simulate('change', { target: { value: 'Ea' } });
+      $input.simulate('change', { target: { value: 'Or' } });
```
ad1992: revert this to `Ea` and change the expected result in line 544 instead

__tests__/reactTags.test.js
```diff
@@ -562,10 +562,10 @@ describe('Test ReactTags', () => {
         })
       );
       const $input = $el.find('.ReactTags__tagInputField');
-      $input.simulate('change', { target: { value: 'Ea' } });
+      $input.simulate('change', { target: { value: 'Or' } });
```
ad1992: revert this to `Ea` and change the expect result in line 568

src/components/ReactTags.js
```diff
@@ -131,13 +131,16 @@ class ReactTags extends Component {
       return this.props.handleFilterSuggestions(query, suggestions);
     }
 
+    const queryIndex = (item) =>
```
ad1992: Add a spec under https://github.com/prakhar1989/react-tags/blob/75fd6ad6a6f9c36586b20cf1afd02fa3b2bca053/__tests__/reactTags.test.js#L421 describe block to check when multiple strings are rendered using partial search, example you can add one more suggestion `orange` in the suggestion list and search using `an` so that `banana` and `orange` both are shown i suggestions.

src/components/ReactTags.js
Gibr: I will talk about contries as example for explanation.
When we use "to" as query for search:
- in my version we get "**To**go, **To**ngo, Equa**to**rial Guinea, **Es**tonia, ...etc."
- in merged variant (>=0) we get "Equa**to**rial Guinea, **Es**tonia, ...etc., and only at the end **To**go, **To**ngo"
My opinion - first variant is more suitable, but the decision is yours.
What do you think?

__tests__/reactTags.test.js
Gibr: Accepted

__tests__/reactTags.test.js
Gibr: Accepted

src/components/ReactTags.js
Gibr: If i will revert previous samples from "Or" to "Ea" it will already shows partial search with result "Pear" and "Peach", isn't it?
Is it still necessary to add more suggestion and test?

Gibr: @ad1992 thank you for your feedback. See my comments for requested changes.

Gibr:
@ad1992 thank you for your feedback. See my comments for requested changes.

src/components/ReactTags.js
ad1992: @Gibr I get your point. I agree with you. In that case can you make two variables, `exactSuggestions` and `partialSuggestions` and then return `exactSuggestions.concat(partialSuggestions)` so its more clear ?

src/components/ReactTags.js
```diff
@@ -131,13 +131,17 @@ class ReactTags extends Component {
...
-      return (
-        item[this.props.labelField]
-          .toLowerCase()
-          .indexOf(query.toLowerCase()) === 0
-      );
+    const queryIndex = (item) =>
+      item[this.props.labelField]
+        .toLowerCase()
+        .indexOf(query.toLowerCase());
+    const exactSuggestions = suggestions.filter((item) => {
```
Gibr: Added `exactSuggestions` and `partialSuggestions` variables with `return exactSuggestions.concat(partialSuggestions);`

__tests__/reactTags.test.js
```diff
@@ -426,9 +426,15 @@ describe('Test ReactTags', () => {
         defaults.suggestions
       );
 
-      $input.simulate('change', { target: { value: 'ea' } });
+      $input.simulate('change', { target: { value: 'or' } });
       expect(ReactTagsInstance.state.suggestions).to.have.members([]);
 
+      $input.simulate('change', { target: { value: 'ea' } });
```
Gibr: Added 'ea' test case for handle partial search behavior. 
'or' test case still here for handle  empty search behavior.

__tests__/reactTags.test.js
```diff
@@ -538,10 +544,10 @@ describe('Test ReactTags', () => {
       );
       const $input = $el.find('.ReactTags__tagInputField');
 
-      $input.simulate('change', { target: { value: 'Ea' } });
+      $input.simulate('change', { target: { value: 'Or' } });
```
Gibr: 'Or' still here because test message 'handles addition when using default suggestions filter' mean there have to be query with result of zero suggestions to add new tag, as i understand.

__tests__/reactTags.test.js
```diff
@@ -562,10 +568,10 @@ describe('Test ReactTags', () => {
         })
       );
       const $input = $el.find('.ReactTags__tagInputField');
-      $input.simulate('change', { target: { value: 'Ea' } });
+      $input.simulate('change', { target: { value: 'Or' } });
```
Gibr: 'Or' still here because test message 'should add tag with custom label field and default suggestion filter' mean there have to be query with result of zero suggestions to add new tag, as i understand.

src/components/ReactTags.js
```diff
@@ -131,13 +131,17 @@ class ReactTags extends Component {
       return this.props.handleFilterSuggestions(query, suggestions);
     }
 
-    return suggestions.filter((item) => {
-      return (
-        item[this.props.labelField]
-          .toLowerCase()
-          .indexOf(query.toLowerCase()) === 0
-      );
+    const queryIndex = (item) =>
```
ad1992: can you move this to an instance function `getQueryIndex` ?

Gibr: done

ad1992:
@Gibr thanks for your contribution

Gibr:
@ad1992 thanks for your feedback

Issue #488, opened 2019-02-08
[Issue] Suggestion refresh after handleAddition execution


LukaszNowakPL:
**Issue**
Suggestions list does not refresh and contains suggestion already taken.

**Steps to reproduce**
- Make suggestion panel open (in my case after clicking on input form - minQueryLength={0}
- Choose one of suggestions provided
- Make suggestions panel open again

**My code**
```
<ReactTags
	tags={tags}
	suggestions={suggestions}
	handleAddition={this.handleAddClick}
	handleDelete={i => this.handleDeleteClick(tags, i)}
	minQueryLength={0}
	autocomplete={true}
	autofocus={false}
	handleFilterSuggestions={this.handleSuggestions}
/>
```

**Explanation**
My `handleAddClick` method modifies `tags` and `suggestions` so first one contains taken element and second one does not (checked by console.log one line before render). Nevertheless component update seems not to picking up new `suggestions` list.

Sneaking to the code, it seems `state.suggestions` is modified only on constructor as well as during `handleChange` execution. This method is executed only by input's `onChange`. In this case `suggestions` list is refreshed only after typing at least one sign into input form.

ad1992:
@LukaszNowakPL I agree with you we should refresh the suggestions. I think We should refresh only when `allowUnique` prop is false. What are your thoughts ?

Gibr:
Maybe it would be better to filter out somehow selected tags from suggestion list during `filteredSuggestions` whithout modifyng suggestions list?

Issue #484, opened 2019-02-05
[Issue] Filtering and updating suggestions


dmossie:
I want to show all suggestions when I click the input field. So I set minQueryLength to 0.

However I want to only show the suggestions that are not included in my current tags. I changed my suggestions prop, but that didn't update react-tags.

I noticed I could use handleFilterSuggestions for this, however it is not called when the input field is first clicked, and is only called on keystroke.

However since I want to only show the suggestions which are not tags with a minQueryLength of zero, this is an issue.

Can someone please help me out? I need to either be able to filter onClick of the suggestion box, or update the suggestions itself

dmossie:
Haven't tested it, but I'm thinking L197 of ReactTags needs something like this:

const suggestions = this.state.suggestions;
if (this.props.suggestions && this.props.minQueryLength && this.props.minQueryLength === 0){
	suggestions = this.filteredSuggestions(query, this.props.suggestions);
}
this.setState({ 
	isFocused: true,
	suggestions: suggestions
});

ad1992:
@dmossie seems like the issue is similar to https://github.com/prakhar1989/react-tags/issues/488 ?

dmossie:
@ad1992 Yeah it looks like the issues we are having are coming from the same problem with the library.

PR #512, opened 2019-03-09
Filter out duplicated tags

Gibr:
This PR partly covers issue #488 and #484 .
It does not provide suggestion list automatically update.
It provides filtering out already selected tags from suggestions depend on `allowUnique` property.
Tests:
At this moment default filter test case with query 'ap' doesn't expect 'apple' because it already in tags list.

### DIFF
--- a/src/components/ReactTags.js
+++ b/src/components/ReactTags.js
@@ -141,11 +141,14 @@ class ReactTags extends Component {
       return this.props.handleFilterSuggestions(query, suggestions);
     }
 
+    const { tags, allowUnique } = this.props;
+    const isTagUnique = (item) =>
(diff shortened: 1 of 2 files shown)


src/components/ReactTags.js
```diff
@@ -147,7 +147,14 @@ class ReactTags extends Component {
     const partialSuggestions = suggestions.filter((item) => {
       return this.getQueryIndex(query, item) > 0;
     });
-    return exactSuggestions.concat(partialSuggestions);
+    let totalSuggestions = exactSuggestions.concat(partialSuggestions);
```
ad1992: can you add a spec for this?

PR #525, opened 2019-04-22
feat(ReactTags): Add option to disable input

### LINKED ISSUE #194: Add option to disable input.

atiqueansari1987:
fix #194


src/components/ReactTags.js
```diff
@@ -440,6 +443,7 @@ class ReactTags extends Component {
           id={inputId}
           maxLength={maxLength}
           value={this.props.inputValue}
+          disabled={disabled}
```
ad1992: as discussed in previous Pull request, these props will not be needed if we pass inputProps as rest props

ad1992:
@atiqueansari1987 any update ?

PR #526, opened 2019-04-22
feat(ReactTags): Add option to on/off input autocomplete

### LINKED ISSUE #461: Can't set browser autocomplete to "off"
The browser's built-in autocomplete functionality is interfering. I need to be able to add the attribute `autocomplete="off"` to the `input`. I haven't seen any way to do this in the documentation and the code doesn't look like a method is provided. Could this please be added?

atiqueansari1987:
fix #461

### DIFF
--- a/src/components/ReactTags.js
+++ b/src/components/ReactTags.js
@@ -70,6 +70,7 @@ class ReactTags extends Component {
     ),
     allowUnique: PropTypes.bool,
     renderSuggestion: PropTypes.func,
+    inputFieldAutocomplete: PropTypes.string,
   };
 
   static defaultProps = {
@@ -90,6 +91,7 @@ class ReactTags extends Component {
     allowUnique: true,
     allowDragDrop: true,
     tags: [],
+    inputFieldAutocomplete: 'on',
   };
 
   constructor(props) {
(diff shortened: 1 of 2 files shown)


src/components/ReactTags.js
```diff
@@ -439,6 +442,7 @@ class ReactTags extends Component {
           name={inputName}
           id={inputId}
           maxLength={maxLength}
+          autoComplete={inputFieldAutocomplete}
```
ad1992: @atiqueansari1987 lets not make it as a prop. instead lets pass it as a rest prop ? in that way these fields need not have a prop, whatever the parent passes in inputProps will be added here as rest props. what do you think ?

atiqueansari1987: Yes, we can have something like this.

```
<input
          ref={(input) => {
            this.textInput = input;
          }}
          className={this.state.classNames.tagInputField}
          type="text"
          placeholder={placeholder}
          aria-label={placeholder}
          onFocus={this.handleFocus}
          onBlur={this.handleBlur}
          onChange={this.handleChange}
          onKeyDown={this.handleKeyDown}
          onPaste={this.handlePaste}
          name={inputName}
          id={inputId}
          value={this.props.inputValue}
          {...inputAttributes}
/>
```

Shall I remove input related props and introduce inputAttributes which will accept an object.

ad1992: yup pls remove input related props and rename `inputAttributed` to `inputProps`.

ad1992:
@atiqueansari1987 any update

PR #532, opened 2019-06-03
Add: api for tag validation

karimsa:
Hi, there!

This PR is meant to add a way to explicitly ignore an attempt to add a tag if the tag is invalid. Currently, one can only avoid adding the tag but the input is still reset - which feels more like a bug to the user than an invalid input. With this change, a `false` return value from `handleAddition` will cause the input reset to not happen.

### DIFF
--- a/src/components/ReactTags.js
+++ b/src/components/ReactTags.js
@@ -339,16 +339,16 @@ class ReactTags extends Component {
     }
 
     // call method to add
-    this.props.handleAddition(tag);
+    if (this.props.handleAddition(tag) !== false) {
+      // reset the state
+      this.setState({
+        query: '',
+        selectionMode: false,
+        selectedIndex: -1,
+      });
 
-    // reset the state
-    this.setState({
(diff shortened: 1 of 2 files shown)


src/components/ReactTags.js
```diff
@@ -339,16 +339,16 @@ class ReactTags extends Component {
...
     // call method to add
-    this.props.handleAddition(tag);
-
-    // reset the state
-    this.setState({
-      query: '',
-      selectionMode: false,
-      selectedIndex: -1,
-    });
+    if (this.props.handleAddition(tag) !== false) {
```
ad1992: instead of returning `false` or `true` can you add a new prop `isTagValid` ? This prop when not passed from parent will return true as default prop and when present will be used to decide if `handleValidation` should be called ?

ad1992:
Can you also make the commit of the pr and title semantic

ad1992:
@karimsa any update ?

PR #548, opened 2019-07-28
Npm audit

JakeSidSmith:
Fixes 1079 vulnerabilities

This should fix several of the open greenkeeper pull requests, and make it much easier to handle dependency audits in the future.

* Un-pin dependencies (allowing patch updates)
* Run `npm audit fix` to automatically fix vulnerabilities

Some additional dev dependencies need manual review: jest, babel-cli, commitizen

Upgrading jest will be quite a bit of work as newer versions use babel 7, and you have some dependencies on deprecated babel plugins.

I would also recommend adding an `.nvmrc` file with a supported version of node for dev. 🙂


package.json
```diff
@@ -36,52 +36,52 @@
     "react-dom": "^16.3.1"
   },
   "dependencies": {
-    "classnames": "2.2.6",
-    "core-js": "2.6.5",
-    "lodash": "4.17.12",
-    "memoize-one": "5.0.1",
-    "prop-types": "15.7.0"
+    "classnames": "^2.2.6",
```
ad1992: remove `^` from all deps as we lock at exact version

JakeSidSmith: Your lock file locks the exact version. If you lock the exact versions for people consuming your library then they can't receive critical vulnerability updates, hence this PR.

Belco90: I'm also interested into get dependencies updated with latest patch versions so I'm really looking forward to merging this PR. @JakeSidSmith is right about locking exact version on package.json and not getting last updated patches (I'm actually having this issue on another repo where I use react-tags as dependency). That's actually the purpose of package-lock.json file, so it's fine to leave then with `^` prefix, and even recommended.

Belco90: Actually, pinning versions on package.json is even dangerous https://medium.com/the-guild/how-should-you-pin-your-npm-dependencies-and-why-2b8d545c7312 so please think about unpinning them and merge this PR please @ad1992

ad1992: @Belco90 thanks for the link, it helped. @JakeSidSmith Thank you for the contribution, let's make the dependencies using `~` just to be little safe ?  with dev deps `^` is fine. What do you think ?

JakeSidSmith: I'm amenable to `~` for dependencies since you don't have any zero-version dependencies. Will update when I get some time. 🙂

ad1992: sure will merge this PR as soon as you make the change. thanks for the contribution  @JakeSidSmith

JakeSidSmith: Hey, @ad1992, sorry it's taken me a while to get that sorted, have been working on other projects. Is it okay to leave the peer dependencies with `^`? It's very hard to match minor versions when it comes to peers, especially with libs like react.

PR #554, opened 2019-09-01
feat: Support textarea

yanggun7201:
I just want to support also `textarea` so that I add `inputType` to support it.
It should be one of the `input` or `textarea`.


src/components/ReactTags.js
```diff
@@ -70,6 +70,7 @@ class ReactTags extends Component {
     ),
     allowUnique: PropTypes.bool,
     renderSuggestion: PropTypes.func,
+    inputType: PropTypes.oneOf(["input", "textarea"]),
```
ad1992: use `singleQuote`

src/components/ReactTags.js
```diff
@@ -90,6 +91,7 @@ class ReactTags extends Component {
     allowUnique: true,
     allowDragDrop: true,
     tags: [],
+    inputType: "input",
```
ad1992: use `singleQuote`

src/components/ReactTags.js
yanggun7201: I fixed it. thanks

src/components/ReactTags.js
yanggun7201: I fixed it. thanks

yanggun7201:
> @yanggun7201 thanks for your contribution, could you make the changes and also add spec for this change

Can I just change `README.md` file? or more files?
Could you please let me know about it?
Thanks.

ad1992:
> > @yanggun7201 thanks for your contribution, could you make the changes and also add spec for this change
> 
> Can I just change `README.md` file? or more files?
> Could you please let me know about it?
> Thanks.

@yanggun7201 can you add a spec for this in `ReactTags.spec.js` file

Issue #559, opened 2019-10-11
Uncaught Invariant Violation: Expected to find a valid target.


francisrod01:
Uncaught Invariant Violation: Expected to find a valid target.

I'm facing this issue when I drag a tag.

![image](https://user-images.githubusercontent.com/4691768/66673624-554d5a80-ec37-11e9-8122-c5c6d32ab1f6.png)

Envs:

- react@16.10.2
- react-dom@16.10.2
- react-tag-input@6.4.1
- react-dnd@5.0.0
- react-dnd-html5-backend@3.0.2

A related tag here: [Uncaught Error: Invariant Violation: Expected to find a valid target](https://github.com/react-dnd/react-dnd/issues/236)

jmfsilva:
DND and a little buggy.
It rarely works and most of the times i'v that error.
Envs:
 - "react": "^16.9.0"
- "react-dom": "^16.9.0"
- "react-dnd": "^5.0.0" (as recommended)
- "react-dnd-html5-backend": "^3.0.2", (as recommended)

jameswyse:
I'm getting the same error. I think it's because the `key` prop isn't stable

Currently it's set to `key={`${tag.id}-${index}`}` in [ReactTags.js:389](https://github.com/prakhar1989/react-tags/blob/master/src/components/ReactTags.js#L389)
Changing it to `key={tag.id}` fixed the problem for me!

tankorsmash:
I'm having the same issue with dragging a tag and throwing the invariant error about not being able to find a target. 

Same as above, except 

    "react": "^16.8.0"
    "@hot-loader/react-dom": "^16.8.0"


Some potential causes is inconsistent ids (maybe based on their position on the array) and keys. We'll see if I can make any progress on my own error.

PR #583, opened 2020-02-25
fix(ReactTags): drag and drop not working

tankorsmash:
As mentioned in an issue, #559, if you drag and drop, React DND throws an invariant issue.

This change makes it so if you pass in a third property to each of your tags, you can have drag and drop. 

I haven't tested any other usecases beyond mine though, so if anyone else wanted to confirm this works for them, that would be appreciated. Meanwhile, the change can be made in your local `node_modules\react-tag-input\dist-modules\components\ReactTags.js#523`

### DIFF
--- a/src/components/ReactTags.js
+++ b/src/components/ReactTags.js
@@ -387,7 +387,7 @@ class ReactTags extends Component {
     return tags.map((tag, index) => {
       return (
         <Tag
-          key={`${tag.id}-${index}`}
+          key={tag.key || tag.id}
           index={index}
           tag={tag}
           labelField={labelField}
--- a/src/components/Tag.js
+++ b/src/components/Tag.js
@@ -54,6 +54,7 @@ Tag.propTypes = {
   tag: PropTypes.shape({
     id: PropTypes.string.isRequired,
     className: PropTypes.string,
+    key: PropTypes.string,
   }),
   moveTag: PropTypes.func,
   removeComponent: PropTypes.func,


tankorsmash:
comment to hope it retriggers the Semantic Pull Request check after I updated the PR title (never used SemPR before)

tankorsmash:
@prakhar1989 is there anything you'd like to see changed in this?

tankorsmash:
Hope you get a chance to take a look at this

src/components/ReactTags.js
```diff
@@ -387,7 +387,7 @@ class ReactTags extends Component {
     return tags.map((tag, index) => {
       return (
         <Tag
-          key={`${tag.id}-${index}`}
+          key={tag.key || tag.id}
```
ad1992, line 390: can you reverse this, `tag.id || tag.key` since id is getting used to check if the tags are unique

src/components/Tag.js
```diff
@@ -54,6 +54,7 @@ Tag.propTypes = {
   tag: PropTypes.shape({
     id: PropTypes.string.isRequired,
     className: PropTypes.string,
+    key: PropTypes.string,
```
cr0ax, line 57: Should this be optional ?

src/components/ReactTags.js
cr0ax: Wouldn't it be better to use `key` since that's how react usually determines uniqueness in general?
Especially if we make it an optional property of each `tag` item

tankorsmash:
@cr0ax  those both look like good ideas. I've stepped away from my repo for a little while. If you're interested in improving my fork, I think its possible to fork mine and resubmit that one. Thank you for looking at this for me!

cr0ax:
@tankorsmash yeah I don't know who the maintainer of this repo is

just looking to get this moving, as I'm also experiencing a drag and drop issue while using this lib

PR #589, opened 2020-04-12
Remove core-js from dependencies

znarf:
Seems it's not used anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,6 @@
   },
   "dependencies": {
     "classnames": "~2.2.6",
-    "core-js": "~2.6.5",
     "lodash": "~4.17.12",
     "prop-types": "~15.7.0"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -37,7 +37,6 @@
   },
   "dependencies": {
     "classnames": "~2.2.6",
-    "core-js": "~2.6.5",
```
ad1992, line 40: can you also update the lock file?

znarf: @ad1992 usually `package-lock.json` should not be committed for libraries, so I suggest to delete it instead.

znarf: @ad1992 updated it anyway in case that was a blocker

ad1992: we commit package.json so that all the contributors have the same set up in local development

ad1992:
@znarf thanks for your contribution.

PR #597, opened 2020-05-26
fix(ReactTags): Drag and drop failing

cr0ax:
Resurrected from another PR

Hopefully fixes issue #559

I commented on [this PR](https://github.com/prakhar1989/react-tags/pull/583/files) and the author quickly closed it

I will reiterate the same points I stated over on that one, and test it out to see if it fixes the issue

### DIFF
--- a/src/components/ReactTags.js
+++ b/src/components/ReactTags.js
@@ -387,7 +387,7 @@ class ReactTags extends Component {
     return tags.map((tag, index) => {
       return (
         <Tag
-          key={`${tag.id}-${index}`}
+          key={tag.key || tag.id}
           index={index}
           tag={tag}
           labelField={labelField}
--- a/src/components/Tag.js
+++ b/src/components/Tag.js
@@ -54,6 +54,7 @@ Tag.propTypes = {
   tag: PropTypes.shape({
     id: PropTypes.string.isRequired,
     className: PropTypes.string,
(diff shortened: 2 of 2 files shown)


src/components/ReactTags.js
```diff
@@ -387,7 +387,7 @@ class ReactTags extends Component {
     return tags.map((tag, index) => {
       return (
         <Tag
-          key={`${tag.id}-${index}`}
+          key={tag.key || tag.id}
```
cr0ax, line 390: I don't believe we should have `tag.id` higher priority than `tag.key` for the following reasons:
1) most users this library are not sending a `key` property anyway (virtually no change for them)
2) react uses the `key` property to determine uniqueness already, so this is more in line with the "react way" of doing things

src/components/Tag.js
```diff
@@ -54,6 +54,7 @@ Tag.propTypes = {
   tag: PropTypes.shape({
     id: PropTypes.string.isRequired,
     className: PropTypes.string,
+    key: PropTypes.string,
```
cr0ax, line 57: this is already optional, omitting it will just default to (something like) the old behavior

cr0ax:
Confirmed this does fix the bug

ad1992:
Thanks for the contribution @cr0ax.

PR #699, opened 2021-02-12
fix(ReactTags): Fix delete error when tags are empty

prateek3255:
When there are no tags left and the user tries to delete tags, we get this error for the line, because the value of `currentTags[index]` will be `undefined`
https://github.com/prakhar1989/react-tags/blob/19faa054bf33e9d355e8f31e7558ccd92c981056/src/components/ReactTags.js#L167

Image 2021-02-12 at 9 37 30 PM

I have added any early return from the `handleDelete` function to avoid this error


ad1992 (approved):

ad1992:
Thanks for the contribution @prateek3255 🎉

PR #752, opened 2021-06-25
fix: redefine new line delimiter

### LINKED ISSUE #721: Edge: pasting multiline text creates multiple tags
## Expected Behavior
Pasted multiline text creates only one tag.

Callback passed to `handleAddition` prop to be called once with a pasted text, e.g `{id: "word1\nword2\nword3", text: "word1\nword2\nword3"}`

## Current Behavior
Pasted multiline text creates multiple tags.

Callback passed to `handleAddition` prop to be called multiple times with a partial pasted text, e.g `{id: "word1", text: "word1"}`, `{id: "\n", text: "\n"}`, `{id: "\nword2", text: "\nword2"}`, `{id: "\nword3", text: "\nword3"}`

## Steps to Reproduce
1. Open React Tags Example (https://stackblitz.com/edit/react-tag-input-1nelrc) in Edge 90 / browserstack.com (Edge 90)
2. Paste multiline text
```
word1

word2



word3
```

pietruszka:
fixes #721 

Fixes problem with pasting multiline text. 
Based on the https://github.com/react-tags/react-tags/blob/master/src/components/constants.js#L2 , enter delimiter is marked as unicode `13` (Carriage return), which is partially true. Enter should be also marked with unicode `10` (Line feed). Please see https://stackoverflow.com/a/1761064
That means on windows/edge it works in partial - it splits a text by `\r`:  `word1\r\nword2` -> `['word1', '\nword2']` but in other systems it does not split text at all.

Tested on Chrome, Safari, Edge, Firefox

### DIFF
--- a/src/components/RemoveComponent.js
+++ b/src/components/RemoveComponent.js
@@ -7,7 +7,7 @@ const RemoveComponent = (props) => {
   const { readOnly, removeComponent, onRemove, className, tag, index } = props;
 
   const onKeydown = (event) => {
-    if (event.keyCode === KEYS.ENTER || event.keyCode === KEYS.SPACE) {
+    if (KEYS.ENTER.indexOf(event.keyCode) !== -1 || event.keyCode === KEYS.SPACE) {
       event.preventDefault();
       event.stopPropagation();
       return;
--- a/src/components/RemoveComponent.js
+++ b/src/components/RemoveComponent.js
@@ -7,7 +7,7 @@ const RemoveComponent = (props) => {
   const { readOnly, removeComponent, onRemove, className, tag, index } = props;
 
   const onKeydown = (event) => {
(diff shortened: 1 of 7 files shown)


ad1992 (approved):
Thanks for fixing this @pietruszka 🎉

src/components/RemoveComponent.js
```diff
@@ -7,7 +7,7 @@ const RemoveComponent = (props) => {
   const { readOnly, removeComponent, onRemove, className, tag, index } = props;
 
   const onKeydown = (event) => {
-    if (event.keyCode === KEYS.ENTER || event.keyCode === KEYS.SPACE) {
+    if (KEYS.ENTER.indexOf(event.keyCode) !== -1 || event.keyCode === KEYS.SPACE) {
```
ad1992, line 10:
```suggestion
    if (KEYS.ENTER.includes(event.keyCode) || event.keyCode === KEYS.SPACE) {
```
for better readability as it returns boolean

PR #760, opened 2021-07-06
fix: Add type to remove button

jdups-cordeco:
I wish to use ReactTags inside a form
To prevent tag's removing when I press on Enter (to submit the form), I do need to set a type to all my buttons.
ReactTags did not set any type to this RemoveComponent's button. So I added it.


ad1992:
Thanks for  the contribution @jdups-cordeco, can you add a snapshot for remove button as well ?

ad1992:
Thanks @jdups-cordeco for the contribution 🎉

PR #828, opened 2022-01-16
chore: change globs in npm format script

christowiz:
The globs used in the `format` script are incorrect for nested files and is logging a "No files matching the pattern were found" error


ad1992:
@christowiz thanks for fixing this!

Issue #500, opened 2019-02-19
increase test coverage


ad1992:
Add missing tests for this library

ad1992:
Sure  @avranju94 please go ahead 👍🏻

Genne23v:
@ad1992  I picked up this ticket and sent a PR to improve test coverage.

PR #880, opened 2022-10-20
test(removecomponent.test): add RemoveComponentTests to improve test …

Genne23v:
I picked up the [issue-500](https://github.com/react-tags/react-tags/issues/500). 
This PR will improve RemoveComponent test coverage. Please let me know if you have any feedback. 
Thank you

### DIFF
--- a/__tests__/RemoveComponent.test.js
+++ b/__tests__/RemoveComponent.test.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
 import { stub } from 'sinon';
 
 import RemoveComponent from '../src/components/RemoveComponent';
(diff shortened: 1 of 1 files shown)


Genne23v:
Codacy recommendation is to use `toHaveLength()` instead. But I found this is not the case. `toHaveLength()` looks for `length` property before comparing according to [Jest documentation](https://jestjs.io/docs/expect#tohavelengthnumber). It seems returning object from the function doesn't have `length` property. Please let me know if you have another recommendation. Thanks

Genne23v:
I think I used the method in a wrong way. Now all checks pass. Please review my PR. Thank you!

Genne23v:
@ad1992 I would appreciate if you could review my PR.

__tests__/RemoveComponent.test.js
```diff
@@ -20,9 +20,47 @@ describe('Test <RemoveComponent/>', () => {
   afterAll(() => {
     onRemoveStub.restore();
   });
+  function mockItem(override) {
```
ad1992, line 23:
```suggestion

  function mockItem(override) {
```
newline added

__tests__/RemoveComponent.test.js
```diff
@@ -20,9 +20,47 @@ describe('Test <RemoveComponent/>', () => {
...
+      onRemove: stub(),
+      index: 0,
+      removeComponent: stub(),
+      tag: {
+        id: 'Tags',
+        key: 'Tags',
+      },
+    };
+    const wrapper = shallow(<RemoveComponent { ...props } />);
+    expect(wrapper.find('button')).toHaveLength(0);
```
ad1992, line 58: not clear what are you trying to test here ? Shouldn't this return a remove component instead ?

ad1992 (commented):
@Genne23v thanks for adding the tests!
Left a comment and then we can merge

__tests__/RemoveComponent.test.js
Genne23v: Will be addressed

__tests__/RemoveComponent.test.js
Genne23v: What I'm trying to test is that there shouldn't be a `button` tag when `removeComponent` option is provided. If it doesn't make sense, I will remove this test. Thanks

ad1992: Instead of checking whether the button is rendered, pass a custom component (eg a span with some text) and check if that is rendered.

Genne23v: Sorry for my lack of understanding, I think the test does not make sense to me. And the coverage is same without this test. Let me remove and push it again. Thanks

ad1992 (approved):
Merging, thanks @Genne23v 🎉

Genne23v:
Thank you! It is my pleasure to contribute.

Issue #596, opened 2020-05-14
Useful callback data for handleInputChange, handleInputFocus, handleInputBlur


agusterodin:
It would be awesome if a reference to the synthetic event itself was provided in the handleInputChange callback, handleInputFocus, and handleInputBlur callback. This would be especially useful for the handleInputChange event (so you could do stuff with event.target)

PR #908, opened 2023-02-10
feat: expose original event on event handlers

joaosamouco:
Hi :wave:

I have one use case where I need to access the original event on `handleInputFocus`. I noticed that more people would like to have this exposed on other event handlers so I added it to every event handler to be consistent.

closes https://github.com/react-tags/react-tags/issues/596


ad1992 (approved):
Thanks @joaosamouco

PR #913, opened 2023-04-11
Update css for remove tag button

vanbeonhv:
I should be `button.ReactTags__remove` instead of `a.ReactTags__remove`. The remove button is `<button>` tag, not `<a>` tag


vanbeonhv (commented):
.

Issue #920, opened 2023-08-09
Add support for max number of tags allowed


timothytong:
When the user has created X number of tags (x defined in a prop), hide the input so that the user can no longer create more tags. Unhide the input once one of the tags has been removed (number of tags < limit)

faizanu94:
This is implemented in https://github.com/react-tags/react-tags/pull/933 and available for review @ad1992

ad1992:
closed by https://github.com/react-tags/react-tags/pull/933

PR #933, opened 2023-12-24
feat: Add support for max number of tags allowed

faizanu94:
This PR adds support for max number of tags through maxTags (optional) prop, reference issue #920

### DIFF
--- a/__tests__/__snapshots__/reactTags.test.js.snap
+++ b/__tests__/__snapshots__/reactTags.test.js.snap
@@ -44,7 +44,6 @@ exports[`Test ReactTags When ClearAll is true should render a clear all button 1
         >
           Clear all
         </button>
-        
       </div>
     </div>
   </div>
--- a/__tests__/__snapshots__/reactTags.test.js.snap
+++ b/__tests__/__snapshots__/reactTags.test.js.snap
@@ -44,6 +44,7 @@ exports[`Test ReactTags When ClearAll is true should render a clear all button 1
         >
           Clear all
         </button>
+        
       </div>
     </div>
   </div>
--- a/__tests__/reactTags.test.js
+++ b/__tests__/reactTags.test.js
@@ -935,7 +935,7 @@ describe('Test ReactTags', () => {
     });
   });
 
-  describe('when maxTags is defined', () => {
+  describe('When maxTags is defined', () => {
     it('should disable adding tags when tag limit reached', () => {
       const tags = [{ id: 'A', text: 'A' }];
       const $el = render(
@@ -955,8 +955,9 @@ describe('Test ReactTags', () => {
         keyCode: ENTER_ARROW_KEY_CODE,
       });
       expect(tags).to.have.length(1);
-      expect($el.getByTestId('tag-limit-message').textContent).to.exist;
-      expect($el.getByTestId('tag-limit-message').textContent).to.equal('tag limit reached');
+      expect($el.getByTestId('error').textContent).to.equal(
+        'Tag limit reached!'
+      );
     });
   });
 });
--- a/__tests__/reactTags.test.js
+++ b/__tests__/reactTags.test.js
@@ -934,4 +934,22 @@ describe('Test ReactTags', () => {
       expect(onClearAllStub.calledOnce).to.be.true;
     });
   });
+
+  describe('When maxTags is defined', () => {
+    it('should render input when maxTags is nor defined', () => {
+      const RenderResult = render(mockItem());
+      expect(RenderResult.queryByTestId('input')).to.not.be.null;
+    });
+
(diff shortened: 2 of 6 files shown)


__tests__/reactTags.test.js
```diff
@@ -934,4 +934,22 @@ describe('Test ReactTags', () => {
...
+    });
+
+    it('should render input when tags are added less than maxTags value', () => {
+      const RenderResult = render(mockItem({ maxTags: 2 }));
+      expect(RenderResult.queryByTestId('input')).to.not.be.null;
+    });
+
+    it('should not render input when tags are added upto maxTags value', () => {
+      const RenderResult = render(mockItem({ maxTags: 1 }));
+      console.log('aa', RenderResult.queryByTestId('input'));
```
ad1992, line 951:
```suggestion
```

removing logs :)

__tests__/reactTags.test.js
```diff
@@ -934,4 +934,22 @@ describe('Test ReactTags', () => {
       expect(onClearAllStub.calledOnce).to.be.true;
     });
   });
+
+  describe('When maxTags is defined', () => {
+    it('should render input when maxTags is nor defined', () => {
```
ad1992, line 939:
```suggestion
    it('should render input when maxTags is not defined', () => {
```

src/components/ReactTags.js
```diff
@@ -501,27 +503,29 @@ class ReactTags extends Component {
...
-          onKeyDown={this.handleKeyDown}
-          onPaste={this.handlePaste}
-          name={inputName}
-          id={inputId}
-          maxLength={maxLength}
-          value={inputValue}
-          data-automation="input"
-          data-testid="input"
-        />
+        {(!maxTags || (maxTags && tags.length < maxTags)) && (
```
ad1992, line 506: lets return null if max tags reached instead for better readability

ad1992 (changes requested):
Thank you for the PR @faizanu94, requested some changes :)

src/components/ReactTags.js
faizanu94: thanks for the suggestion :)

src/components/ReactTags.js
```diff
@@ -501,27 +503,29 @@ class ReactTags extends Component {
...
-          onKeyDown={this.handleKeyDown}
-          onPaste={this.handlePaste}
-          name={inputName}
-          id={inputId}
-          maxLength={maxLength}
-          value={inputValue}
-          data-automation="input"
-          data-testid="input"
-        />
+        {maxTags && tags.length >= maxTags ? null : (
```
ad1992, line 506: @faizanu94 lets do early return `null` since nothing will be rendered when max tags reached

ad1992 (changes requested):

src/components/ReactTags.js
faizanu94: @ad1992 it appears that an early return won't result in any rendering. Our objective here is to only make the `input` tag conditional. The rest of the JSX, including the Suggestions component, will be rendered regardless of this condition. Please correct me if I'm mistaken

ad1992: Got it, yes you are right,how about disabling the input instead of hiding it ? that could be better UI and we can show an message that max tags reached below the input as well

faizanu94: Agreed, this approach would enhance the UI. However, I'm just curious about the existing `disabled` attribute as defined in the `inputProps` of the library (https://github.com/react-tags/react-tags?tab=readme-ov-file#inputprops). There's a possibility that this could introduce edge cases that may conflict with our requirements (for example: when user explicitly passes `disabled={false}` prop). In an attempt to implement this, I encountered a failing test case (https://github.com/react-tags/react-tags/blob/master/__tests__/reactTags.test.js#L784), which prompted me to reconsider the feasibility of this integration. I would appreciate your insights on this matter

ad1992: @faizanu94 thats a good point, in that case how about disable adding the tags and show an error message below the input? (the input field will be visible)

faizanu94: @ad1992 just added this handling and tried to cover all possible scenarios (clearAll, delete tag, paste etc). Unit test also updated. Awaiting feedback, thanks

__tests__/__snapshots__/reactTags.test.js.snap
```diff
@@ -44,6 +44,7 @@ exports[`Test ReactTags When ClearAll is true should render a clear all button 1
         >
           Clear all
         </button>
+
```
ad1992, line 47:
```suggestion
```

ad1992 (approved):

ad1992:
Merging, thanks @faizanu94 ✨

Issue #930, opened 2023-11-24
allowUnique prop allows to add non-unique values


d-batmanov:
Hi everyone! 
I've run into the problem that in version 6.8.1 `allowUnique` prop is not working correctly

**Expected result:**
When I insert a set of values separated by a comma, only unique values should be added

**Test strings:**
1. `test,test,test` <--- Only one `test` value is added. Correct result
2. `test, test, test` <--- There is a space after the comma, the first and second values are added, and the third value is ignored because it is equal to the second value. `'test' !== ' test'` 

**Expected result:**
No matter how many spaces there are after the delimiter character, the value should still be considered unique

[Demo project](https://stackblitz.com/edit/react-tag-input-z5fach?file=index.js)

PR #939, opened 2024-02-10
fix: trim the tags before adding

ad1992:
closes https://github.com/react-tags/react-tags/issues/930


__tests__/__snapshots__/reactTags.test.js.snap
```diff
@@ -44,6 +44,7 @@ exports[`Test ReactTags When ClearAll is true should render a clear all button 1
         >
           Clear all
         </button>
+
```
ad1992, line 47:
```suggestion
```

PR #976, opened 2024-05-30
build: Change export order

RobertStanyon:
Make default export last in package.json.

Certain build tools (microbundle) require the default export to be last in the exports object.  See e.g. https://stackoverflow.com/questions/76127288/module-not-found-error-default-condition-should-be-last-one.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,10 +9,10 @@
   "exports": {
     ".": {
       "import": "./dist/index.js",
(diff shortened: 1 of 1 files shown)


ad1992:
Ahh I wasn't able to recal the reason for which I had moved the default exports to the last in of the other libraries, thanks @RobertStanyon

ad1992 (approved):

Issue #979, opened 2024-07-03
Incorrect classNames typedef


chasemcdo:
Hey team!

I am looking to use this package in one of my projects, but am having trouble applying classes to it. I am following the documentation and have something like:
```javascript
<ReactTags
  classNames={{
    root: "bg-red-500",
    rootFocused: "bg-red-500",
    selected: "bg-red-500",
    selectedTag: "bg-red-500",
    selectedTagName: "bg-red-500",
    search: "bg-red-500",
    searchInput: "bg-red-500",
    suggestions: "bg-red-500",
    suggestionActive: "bg-red-500",
    suggestionDisabled: "bg-red-500",
  }}
/>
```

Where the Tailwind Class `bg-red-500` is being used as an obvious placeholder for relevant classes; however, the only classes that seem to impact the appearance are those applied to `selected` and on the `Tag` objects themselves. Am I simply misunderstanding how these should be impacting my UI?

To further elaborate, if I try to add any other attributes to classNames, I'll receive type errors as I am running in a typed node environment, but it seems the classNames type def is missing relevant ones while containing unused ones:

https://github.com/react-tags/react-tags/blob/cb3c1c073034be1e93a7a0d35d9707a25bf6e1ea/src/index.tsx#L143-L154

Thanks!

chasemcdo:
Created a [small PR](https://github.com/react-tags/react-tags/pull/980/files), which is likely missing some classes, but updates it to align with the constants definitions here:

https://github.com/react-tags/react-tags/blob/cb3c1c073034be1e93a7a0d35d9707a25bf6e1ea/src/components/constants.ts#L24-L36

chasemcdo:
Workaround for anyone else hitting a similar issue is to simply typecast the classNames input as any or you can type cast it as the correct version from above.

ie
```typescript
<ReactTags
  classNames={{
    selected: "bg-red-500",
    tagInputField: "rounded-lg text-sm px-2 py-1",
  } as any}
/>
```

ad1992:
closed by #980

PR #980, opened 2024-07-03
fix(classNames): update the className attributes so its correctly typed and passed to the DOM

chasemcdo:
Details found in atttached issue:
- #979

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -141,16 +141,17 @@ export interface ReactTagsWrapperProps {
    * CSS class names for the component.
    */
   classNames?: {
-    root: string;
-    rootFocused: string;
-    selected: string;
-    selectedTag: string;
-    selectedTagName: string;
-    search: string;
-    searchInput: string;
-    suggestions: string;
-    suggestionActive: string;
-    suggestionDisabled: string;
+    tags: string,
+    tagInput: string,
+    tagInputField: string,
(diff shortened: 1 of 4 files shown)


src/index.tsx
```diff
@@ -141,16 +141,17 @@ export interface ReactTagsWrapperProps {
...
+    tagInput: string,
+    tagInputField: string,
+    selected: string,
+    tag: string,
+    remove: string,
+    suggestions: string,
+    activeSuggestion: string,
+    editTagInput: string,
+    editTagInputField: string,
+    clearAll: string,
```
ad1992, line 154: Hey @chasemcdo thank you for the fix!
I was traveling so I couldn't check this one. I am back now.

The fix looks good, thanks for catching this! Even though this has been there for ages but none of them would have worked as these classes weren't used anywhere in the codebase

can you write a test for this? You can just verify if the class names are passed correctly to the dom element (taking snapshot)

ad1992 (changes requested):

src/index.tsx
chasemcdo: Added the test and also updated the README while I was at it, since we are running a yarn lockfile for dev, but the readme directed people to use npm.

ad1992 (approved):
Looking good! Thanks @chasemcdo 🚀

PR #990, opened 2024-08-14
fix(exports): properly export tag type

chasemcdo:
This PR adds the export of the `Tag` type to the index file to allow users to import it via `import { Tag } from "react-tag-input";`

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -303,3 +303,4 @@ const WithContext = ({ ...props }: ReactTagsWrapperProps) => (
 export { WithContext };
 export { ReactTagsWrapper as WithOutContext };
 export { KEYS, SEPARATORS };
+export { Tag };


ad1992:
Hi @chasemcdo , thank you for the PR, may I know is there a specific usecase, you want to import Tag component specifically?

chasemcdo:
> Hi @chasemcdo , thank you for the PR, may I know is there a specific usecase, you want to import Tag component specifically?

This is for the base use case where you want type definitions on the tags you pass to the tag component.

at present you need to import it directly from the complete path, while this enables simpler importing of the Tag type

chasemcdo:
Looks like this issue has been addressed in a later version of the package!

chasemcdo:
Actually I lied it is still an issue. To elaborate on what the problem is:
1. In the docs you are required to use the `Tag` interface for typing in TS
2. This type is imported into the index file, but never exported, so if you try to add `import { Tag } from "react-tag-input"` you will get a TS error `Module '"react-tag-input"' declares 'Tag' locally, but it is not exported`
3. Solution for this is exporting `Tag` from the index file as done in this PR.

Presently I have to either redeclare the type in my project or import from `node_modules/react-tag-input/types/components/SingleTag`, which are both non-ideal scenarios.

chasemcdo:
@ad1992

ronjb04:
facing the same issue

sakgoyal:
any update on this?

chasemcdo:
> any update on this?

Not sure what is blocking merging of this myself. @ad1992

ad1992 (approved):
Apologies for the delay!
Thanks for the fix @chasemcdo. Merging!

sakgoyal:
Awesome! Still needs a release for this though

ad1992:
The change has been released in [v6.10.4](https://github.com/react-tags/react-tags/releases/tag/v6.10.4)

Issue #988, opened 2024-08-13
An Extra Tag is Created for the Last Letter When Typing in Korean


ABlued:
When typing in Korean, an extra tag is created for the last letter.

https://github.com/user-attachments/assets/2f542235-250f-4666-8f0d-17263cd0fca6

I believe the issue is likely caused by the onKeyDown event. Changing the event handler from `onKeyDown` to `onKeyUp` resolves the bug, 

```
src/components/ReactTag.tsx 512~532
  <input
        {...inputProps}
        ref={(input) => {
          textInput.current = input;
        }}
        className={allClassNames.tagInputField}
        type="text"
        placeholder={placeholder}
        aria-label={placeholder}
        onFocus={handleFocus}
        onBlur={handleBlur}
        onChange={handleChange}
        onKeyUp={handleKeyDown} // onKeyDown -> onKeyUp
        onPaste={handlePaste}
        name={inputName}
        id={inputId}
        maxLength={maxLength}
        value={inputValue}
        data-automation="input"
        data-testid="input"
      />
```

Results after modifying the code

https://github.com/user-attachments/assets/f25163c3-bdc3-4a42-ab83-d6ad27bcdf0d

but it also introduces side effects that break many of the tests.🥲
Can you solve the bug for this?

ad1992:
Hi @ABlued , thank you for reporting this, will soon look into this.

shrilakshmishastry:
@ad1992 can I inspect more on this?

ad1992:
> @ad1992 can I inspect more on this?

Sure, please go ahead @shrilakshmishastry

shrilakshmishastry:
this issue is similar to https://github.com/vuejs/vue/issues/10277

ad1992:
@ABlued could you confirm if its fixed in https://react-tags-git-fork-shrilakshmishastry-fix-on-4626e5-react-tags.vercel.app/ ?

ABlued:
@ad1992 The issue has been completely resolved! You solved it with `nativeEvent.isComposing`. Thank you for sharing the solution! 👍

ad1992:
> @ad1992 The issue has been completely resolved! You solved it with `nativeEvent.isComposing`. Thank you for sharing the solution! 👍

Awesome! Thanks for testing it out confirming the fix. Thanks to @shrilakshmishastry for pushing the fix ✨

PR #995, opened 2024-09-02
fix: Korean input trigger keydown event twice

shrilakshmishastry:
>[!NOTE]
> The solution needs to be tested by a proper Korean language user. Being a non-Korean lang user, I can change the keyboard and type input and test, but I will not be able to say whether what I tested is proper according to the Korean language, 

## Issue

Korean and Japanese languages characters are formed using the compositions of keys. 

From GPT:

Composition-based Input:
Korean, like many East Asian languages, often uses an Input Method Editor (IME) for text entry. This involves a composition process where multiple keystrokes may be required to form a single character.

Challenges:

1. The value of the input field may not be final until the composition is complete.
2. Key events may fire multiple times for a single Korean character.
3. The keyCode or which properties may not accurately represent the Korean character being input.

The issue we were facing is that the keys were still composing when we were handling `onKeyDown ` event. By the time the composition is done, it is used to trigger one more time enter and last character. 

can read more [here](https://github.com/vuejs/vue/issues/10277)

## Fix

Add a composing check-in on the KeyDown event to handle the next stages. Though Key events might fire multiple times, this makes sure, that until all composing is done enter is not triggered .
 
 
 Why the handling is added in `onKeyDown` event not on `onChange`?
 - we are concentrating on `enter key` which is the terminating key. It's not considered as value in the input tag. 

This fixes https://github.com/react-tags/react-tags/issues/988

## SCREEN RECORDINGS

English lang
https://github.com/user-attachments/assets/fe4060cf-d51f-4fa5-9efd-7f99a20f575f

Korean lang
https://github.com/user-attachments/assets/80ba9760-6184-4faa-8a76-4189ebc8901b

### DIFF
--- a/src/components/ReactTags.tsx
+++ b/src/components/ReactTags.tsx
@@ -1,4 +1,11 @@
-import React, { useEffect, createRef, useRef, useState, Fragment } from 'react';
+import React, {
+  useEffect,
+  createRef,
+  useRef,
+  useState,
+  Fragment,
+  useMemo,
+} from 'react';
 import { uniq } from 'lodash-es';
 import ClearAllTags from './ClearAllTags';
 import Suggestions from './Suggestions';
@@ -16,6 +23,7 @@ import {
   ERRORS,
 } from './constants';
 import type { ReactTagsWrapperProps } from '../';
+import Fuzzy from 'fuzzify';
 
 /**
  * Props for the ReactTags component.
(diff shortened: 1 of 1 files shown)


shrilakshmishastry:
Hey @ad1992 , here is the fix for https://github.com/react-tags/react-tags/issues/988. How to ask the bug reporter to try this out and check whether this works ?

ad1992:
@shrilakshmishastry looks like this has fuzzy PR changes as well, can you only push the changes related to the fix?

ad1992:
@shrilakshmishastry please leave a comment about why this PR is closed

shrilakshmishastry:
The PR is closed as the branch is cut out from non master branch

PR #996, opened 2024-09-02
fix: Korean input trigger keydown event twice

shrilakshmishastry:
>[!NOTE]
> The solution needs to be tested by a proper Korean language user. Being a non-Korean language user, I can change the keyboard type and type input and test, but I will not be able to say whether what I tested is proper according to the Korean language, 

## Issue

Korean and Japanese languages characters are formed using the compositions of keys. 

From GPT:

Composition-based Input:
Korean, like many East Asian languages, often uses an Input Method Editor (IME) for text entry. This involves a composition process where multiple keystrokes may be required to form a single character.

Challenges:

1. The value of the input field may not be final until the composition is complete.
2. Key events may fire multiple times for a single Korean character.
3. The keyCode or which properties may not accurately represent the Korean character being input.

The issue we were facing is that the keys were still composing when we were handling `onKeyDown ` event. By the time the composition is done, it is used to trigger one more time enter and last character. 

can read more [here](https://github.com/vuejs/vue/issues/10277)

## Fix

Add a composing check-in on the KeyDown event to handle the next stages. Though Key events might fire multiple times, this makes sure, that until all composing is done enter is not triggered .
 
 
 Why the handling is added on `onKeyDown` event not on `onChange`?
 - we are concentrating on `enter key` which is the terminating key. It's not considered as value in the input tag. 

This fixes https://github.com/react-tags/react-tags/issues/988

## SCREEN RECORDINGS

English lang
https://github.com/user-attachments/assets/fe4060cf-d51f-4fa5-9efd-7f99a20f575f

Korean lang
https://github.com/user-attachments/assets/80ba9760-6184-4faa-8a76-4189ebc8901b

### DIFF
--- a/__tests__/ReactTags.test.tsx
+++ b/__tests__/ReactTags.test.tsx
@@ -1,5 +1,6 @@
-import React from 'react';
+import React, { Children } from 'react';
 import { expect } from 'chai';
+
 import { spy, stub, createSandbox } from 'sinon';
 
 import { WithContext as ReactTags } from '../src/index';
@@ -7,6 +8,8 @@ import { WithContext as ReactTags } from '../src/index';
 import { KEYS, SEPARATORS } from '../src/components/constants';
 import { fireEvent, render, screen } from '@testing-library/react';
 import type { Tag } from '../src/components/SingleTag';
+
+
 
 /* eslint-disable no-console */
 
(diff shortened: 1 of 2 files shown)


shrilakshmishastry:
Hey @ad1992 👋 , here is the fix for https://github.com/react-tags/react-tags/issues/988. How to ask the bug reporter to try this out and check whether this works ?

ad1992:
Thats an interesting finding @shrilakshmishastry 
can you add a testcase as well ?

ad1992:
@shrilakshmishastry also looks like this branch is behind master, could you merge with latest master so the preview is fixed as well

shrilakshmishastry:
> Thats an interesting finding @shrilakshmishastry ✨ can you add a testcase as well ?

Hey @ad1992 👋 here is the commit [test: add test case for multi-byte character language tags](https://github.com/react-tags/react-tags/pull/996/commits/dfdc730d4a8eb01ae1c10f4e27698a47e439f8fb)). I have added the test case for multi byte languages

shrilakshmishastry:
> @shrilakshmishastry also looks like this branch is behind master, could you merge with latest master so the preview is fixed as well

when I tried to take a pull updating my fork with original repo, I got everything up to date

ad1992:
> > @shrilakshmishastry also looks like this branch is behind master, could you merge with latest master so the preview is fixed as well
> 
> when I tried to take a pull updating my fork with original repo, I got everything up to date

Interesting, probably you didn't fetch the latest master in upstream hence it didn't pull the changes.
No worries I will push it now

ad1992 (approved):
Merging! Thanks @shrilakshmishastry ❤️

Issue #983, opened 2024-07-21
Unable to specific class names to a subset of elements for the classNames prop following recent TypeScript changes


khughitt:
Greetings!

After updating to the most recent version of react-tags (6.10.3), I started getting running into issues with the `classNames` when only a subset of the elements are specified, e.g.:

```
# used to work..
classNames={{ 
  tagInputField: "form-control" ,
  remove: "btn btn-primary py-1 m-2",
  tag: 'react-tag'
}}
```

Now:

![2024-07-21-134439_1032x198_scrot](https://github.com/user-attachments/assets/4e5ab1a2-180e-4624-836d-98946de917ad)

My first thought was to just put empty strings for the values I didn't want to override, but this leads to different errors due to react-tag's expectation of non-empty classes:

```
classNames={{
  tags: '',
  tagInputField: "form-control" ,
  remove: "btn btn-primary py-1 m-2",
  tag: 'react-tag', // adds class instead of replaces class as desired...
  tagInput: '',
  selected: '',
  suggestions: '',
  activeSuggestion: '',
  editTagInput: '',
  editTagInputField: '',
  clearAll: '',
}}
```

Which causes problems when attempting to select an option via:

```
# Leads to: "Unexpected Application Error! Element.querySelector: '.' is not a valid selector"
const activeSuggestion = suggestionsContainerRef.current.querySelector(
   `.${classNames.activeSuggestion}`
);
```

A work-around is to simply copy the default class names over and include them in the prop.

But the behavior should probably be changed so that classnames can be specified for only a subset of the elements, which is probably what most users want.

Unrelated to this issue, but it could also be useful, if it's not too difficult, to allow the user to indicate whether they want  custom class names provided to be _added_ to or _replace_ the default ones.. Previously, I think they were always added, and now it seems like they will always replace them.

Thanks for all your work on this!

Cheers,
Keith

ad1992:
Thanks for reporting this @khughitt, I will soon look into it

PR #997, opened 2024-09-14
fix: type issue in classNames prop

shrilakshmishastry:
## Issue

The className property as an object was optional but when the className is provided it was required to provide all the properties declared or needed in className. 

From chatGPT the more explanation 

```
classNames?: {
  tags: string,
  tagInput: string,
  tagInputField: string,
  selected: string,
  tag: string,
  remove: string,
  suggestions: string,
  activeSuggestion: string,
  editTagInput: string,
  editTagInputField: string,
  clearAll: string,
}
```
The ? here means that classNames itself is optional. You can choose to provide or omit this property.
 Each property inside the classNames object is required if the classNames object is provided. You must include all these properties with corresponding string values if you use the classNames object.
 

## Solution : make each property optional
 
 ```
 
 classNames?: {
  tags?: string,
  tagInput?: string,
  tagInputField?: string,
  selected?: string,
  tag?: string,
  remove?: string,
  suggestions?: string,
  activeSuggestion?: string,
  editTagInput?: string,
  editTagInputField?: string,
  clearAll?: string,
};

 ```
 
 
The ? after classNames means that the classNames property itself is optional. When using this type, you might not provide the classNames property at all.
tags?: string and others: Each property inside the classNames object is also optional. This means that you can provide an object with some, all, or none of these properties.

This fixes https://github.com/react-tags/react-tags/issues/983

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -141,17 +141,17 @@ export interface ReactTagsWrapperProps {
    * CSS class names for the component.
    */
   classNames?: {
-    tags: string,
-    tagInput: string,
-    tagInputField: string,
-    selected: string,
-    tag: string,
-    remove: string,
-    suggestions: string,
-    activeSuggestion: string,
-    editTagInput: string,
-    editTagInputField: string,
-    clearAll: string,
+    tags?: string,
+    tagInput?: string,
+    tagInputField?: string,
+    selected?: string,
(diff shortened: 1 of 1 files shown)


ad1992 (approved):
Thanks for the fix @shrilakshmishastry ✨

Issue #999, opened 2024-09-25
Accessibility issues


shrilakshmishastry:
Hey @ad1992 👋  Hope you're doing good. Here are the accessibility issues I have noticed in react-tags . 

- tags are not reachable to edit using the keyboard
- suggestions are not screen reader accessible, when you move across the suggestions VO doesn’t announce it
- Clear all buttons doesn’t explain what to clear. it feels like I am trying to clear texts in input but actually, it does is clear tags
- screen reader should announce the suggestions when the suggestion is shown
- should provide flexibility to add a label to input tag
- proper aria label for tags and nested interaction handle
- when tags are read only do we need to make tags keyboard accessible? most probably not, when they are read-only they will behave as normal text. This should be considered when fixing keyboard accessibility in tags

I want to pick all of these and fix them. Can you please assign me this issue?

ad1992:
@shrilakshmishastry really appreciate your effort in listing the issues. Please go ahead and fix it ❤️
As per readonly tags, yes they should be just keyboard accessible as they are regular text when readonly

shrilakshmishastry:
Hey @ad1992 what do you mean by keyboard accessible? 

tags shouldn't be a tab/focusable element right when read-only?

ad1992:
> Hey @ad1992 what do you mean by keyboard accessible?
> 
> tags shouldn't be a tab/focusable element right when read-only?

Nope, in readonly its just rendering the tags in `div` so individual tags won't be focussible, however it should be announced via SR

shrilakshmishastry:
Hey @ad1992 👋  hope you're doing well. I need your opinion on this 

- nested interaction in tags

The tag as a whole is keyboard-reachable and activates when clicking enter. On active it behaves as an input tag. When active close icon will be hidden. 

The close icon is also in the tab index sequence when the whole tag acts as a clickable. 

I guess the tag should be a button rather than a span.

shrilakshmishastry:
Hey @ad1992 , here is the solution for the above issues that I am planning to implement. Can you please give me a correction?

Solutions:

- Clear all buttons don’t explain what to clear. it feels like I am trying to clear texts in input but actually, it does clear tags

Give a prop to provide an alt tag to clear all CTA in case CTA is provided. If not give the default alt tag as “clear all chosen tags”

- should provide flexibility to add labels to input tag
    
    This is the label for the main input tag. Users don’t know what they are inputting for. Should make label compulsory. 
    
- proper aria label for tags

[...]

PR #1000, opened 2024-10-01
fix: accessibility issue in clear all cta

shrilakshmishastry:
## Fix

**Issue:** Clear all buttons don’t explain what to clear. it feels like I am trying to clear texts in input but actually, it does clear tags

**solution** : Give a prop to provide an alt tag to clear all CTA in case CTA is provided. If not give the default alt tag as “clear all chosen tags”

accessibility tree
| then | now |
|--|--|
|Screenshot 2024-10-01 at 8 19 04 PM | image|

screen reader announcement when clear all CTA is being focused 

Screenshot 2024-10-01 at 8 41 28 PM

This fixes first issue in https://github.com/react-tags/react-tags/issues/999

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ Option | Type | Default | Description
 |[`clearAll`](#clearAll) | `boolean` | `false` | Implies whether 'clear all' button should be shown.
 |[`onClearAll`](#onClearAll) | `Function` |  | This callback if present is triggered when clear all button is clicked.
 | [`maxTags`](#maxTags) | `number` | | The maximum count of tags to be added
+| [`ariaAttrs`](#ariaAttrs) | Object |`{clearAllLabel: "clear all tags"}` | An object used to pass custom accessibility attributes 
 
 ## Styling
 `<ReactTags>` does not come up with any styles. However, it is very easy to customize the look of the component the way you want it. By default, the component provides the following classes with which you can style -
@@ -699,4 +700,9 @@ This callback is if present is triggered when "clear all" button is clicked. You
 
 ### maxTags
 This prop specifies the maximum count of tags to be added. Incase the tags exceed, error will show up to convey the maximum tag limit has reached.
+
+### ariaAttrs
+
+An object used to pass custom accessibility attributes (ARIA properties) to enhance screen reader support and improve accessibility. For example, clearAllLabel defines the accessible label for a "clear all tags" button, providing context for assistive technologies.
+
 
--- a/README.md
+++ b/README.md
@@ -207,7 +207,7 @@ Option | Type | Default | Description
 |[`clearAll`](#clearAll) | `boolean` | `false` | Implies whether 'clear all' button should be shown.
 |[`onClearAll`](#onClearAll) | `Function` |  | This callback if present is triggered when clear all button is clicked.
 | [`maxTags`](#maxTags) | `number` | | The maximum count of tags to be added
-| [`ariaAttrs`](#ariaAttrs) | Object |`{clearAllLabel: "clear all tags"}` | An object used to pass custom accessibility attributes 
+| [`ariaAttrs`](#ariaAttrs) | Object |`{clearAllLabel: "clear all tags"}` | An object containing custom ARIA attributes to enhance screen reader support and improve accessibility
 
 ## Styling
 `<ReactTags>` does not come up with any styles. However, it is very easy to customize the look of the component the way you want it. By default, the component provides the following classes with which you can style -
@@ -703,6 +703,11 @@ This prop specifies the maximum count of tags to be added. Incase the tags excee
 
 ### ariaAttrs
 
-An object used to pass custom accessibility attributes (ARIA properties) to enhance screen reader support and improve accessibility. For example, clearAllLabel defines the accessible label for a "clear all tags" button, providing context for assistive technologies.
+An object containing custom ARIA attributes to enhance screen reader support and improve accessibility. 
+It accepts the below attributes
+```js
+{
+  clearAllLabel?: string // Optional accessible label for a "clear all tags" button improving accessibility by providing a clear, descriptive action label.
+}
 
 
--- a/__tests__/ReactTags.test.tsx
+++ b/__tests__/ReactTags.test.tsx
@@ -1323,7 +1323,7 @@ describe('Test ReactTags', () => {
 
     })
 
-      it('should have  value for aria-label when label is provided', () => {
+      it('should have value for aria-label when label is provided', () => {
       const tags = render(
         mockItem({
           clearAll: true,
--- a/__tests__/ReactTags.test.tsx
+++ b/__tests__/ReactTags.test.tsx
@@ -1362,7 +1362,7 @@ describe('Test ReactTags', () => {
             Clear all
           </button>
       `);
-      tags.debug();
+      
     });
 
     it('should trigger "onClearAll" callback if present when clear all button is clicked', () => {
--- a/__tests__/ReactTags.test.tsx
+++ b/__tests__/ReactTags.test.tsx
@@ -1304,20 +1304,65 @@ describe('Test ReactTags', () => {
   });
 
   describe('When ClearAll is true', () => {
+    it('should have default value for aria-label when label is not provided', () => {
+      const tags = render(
+        mockItem({
+          clearAll: true,
+        })
+      )
+
+       jestExpect(tags.container.querySelector('.ReactTags__clearAll'))
+        .toMatchInlineSnapshot(`
+        <button
+          aria-label="clear all tags"
+          class="ReactTags__clearAll"
+        >
+          Clear all
+        </button>
+      `);
+
+    })
+
+      it('should have  value for aria-label when label is provided', () => {
+      const tags = render(
+        mockItem({
+          clearAll: true,
+          ariaAttrs: {
+            clearAllLabel: 'Clear all selected tags'
+          }
+        })
+      )
+
+       jestExpect(tags.container.querySelector('.ReactTags__clearAll'))
+        .toMatchInlineSnapshot(`
+        <button
+          aria-label="Clear all selected tags"
+          class="ReactTags__clearAll"
+        >
+          Clear all
+        </button>
+      `);
+
+    })
+
     it('should render a clear all button', () => {
       const tags = render(
         mockItem({
           clearAll: true,
         })
       );
-      jestExpect(tags.container.querySelector('.ReactTags__clearAll'))
+      
+      
+      jestExpect(tags.container.querySelector('.ReactTags__clearAll'))      
         .toMatchInlineSnapshot(`
-        <button
-          class="ReactTags__clearAll"
-        >
-          Clear all
-        </button>
+          <button
+            aria-label="clear all tags"
+            class="ReactTags__clearAll"
+          >
+            Clear all
+          </button>
       `);
+      tags.debug();
     });
 
     it('should trigger "onClearAll" callback if present when clear all button is clicked', () => {
--- a/src/components/ClearAllTags.tsx
+++ b/src/components/ClearAllTags.tsx
@@ -4,11 +4,11 @@ const ClearAllTags = (props: {
   classNames: {
     clearAll: string,
   },
-  accessibilityLabel?: string,
+  ariaLabel?: string,
   onClick: () => void,
 }) => {
   return (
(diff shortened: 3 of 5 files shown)


src/index.tsx
```diff
@@ -197,6 +197,10 @@ export interface ReactTagsWrapperProps {
    * Whether to display a button to clear all the tags.
    */
   clearAll?: boolean;
+  /**
+   * accessibility la
```
johnie, line 201: the JSDoc is incomplete

src/components/ReactTags.tsx
```diff
@@ -41,6 +41,7 @@ type ReactTagsProps = ReactTagsWrapperProps & {
   inputProps: { [key: string]: string };
   editable: boolean;
   clearAll: boolean;
+  accessibilityLabelForClearAll?: string;
```
johnie, line 44: **Type Safety:**
You introduced `accessibilityLabelForClearAll` in the props. Since this is user-facing and can impact the accessibility of the application, consider adding stricter type validation for this prop (e.g., explicitly stating it's a required string).

src/index.tsx
```diff
@@ -197,6 +197,10 @@ export interface ReactTagsWrapperProps {
...
+   */
+  accessibilityLabelForClearAll?: string;
```
johnie, line 203: The prop `accessibilityLabelForClearAll` is descriptive, but a shorter name like `clearAllLabel` could make the code more concise and still clear. This is subjective, but shorter names tend to improve readability.

johnie (commented):
There are a few areas where improvements can be made for better readability, maintainability, and consistency.

While this patch doesn’t directly touch the testing layer, it's critical to ensure that the new accessibility feature is covered in tests, specifically ensuring that `aria-label` is correctly rendered for screen readers. A quick addition of a test case for this would ensure regressions don’t happen.

src/index.tsx
shrilakshmishastry: my bad , I thought jsdoc is added looking at auto-complete, will add it

src/index.tsx
shrilakshmishastry: my bad , I thought jsdoc is added looking at auto-complete, will add it

src/components/ReactTags.tsx
shrilakshmishastry: Since  I have given a default value, https://github.com/react-tags/react-tags/blob/e37e8d59b075ec3e50cd60fffcccf2b8b706d152/src/index.tsx#L235 making it a required prop is not necessary I guess.

src/index.tsx
shrilakshmishastry: thanks for the suggestion. My concern is `clearAllLabel` might give the impression to the user that this value is used inside the button as text. What do you think?

src/index.tsx
shrilakshmishastry: added now https://github.com/react-tags/react-tags/blob/f07fc68cfbba2011073e9b33bcf2f5802b8518a8/src/index.tsx#L201

src/components/ClearAllTags.tsx
```diff
@@ -4,10 +4,11 @@ const ClearAllTags = (props: {
   classNames: {
     clearAll: string,
   },
+  accessibilityLabel: string,
   onClick: () => void,
 }) => {
   return (
-    <button className={props.classNames.clearAll} onClick={props.onClick}>
+    <button aria-label={props.accessibilityLabel} className={props.classNames.clearAll} onClick={props.onClick}>
```
ad1992, line 11: I think we don't need a prop for this, and instead just add a descriptive aria label instead. If the user wants to override it later (if we get any such request) then we can consider supporting customization. But for now adding a hard-coded aria label should be good to go.
Let me know what you think

ad1992 (changes requested):
Thanks for the PR @shrilakshmishastry

src/components/ClearAllTags.tsx
shrilakshmishastry: Your comment makes sense, but a hard-coded aria label gives the information to assistive technology users that, on clicking CTA we are going to perform some task. But they will not have context what tags are these? What are they clearing all that stuff, even if we give a good descriptive hardcoded value 

They feel something like this  https://inaccessible.gallery/form-labels/

There is a saying 
 <blockquote>Accessibility isn't optional: It's integral to website performance and user experience</blockquote>

 To have a better accessible CTA, I feel we should consider taking the aria label from the user.

ad1992: The aria label - **"Clear all tags"** clearly conveys that it would be clearing all the tags. It doesn't clear some specific tags (which you can select to remove) so there isn't any extra info needed to convey about which specific tags will be removed since all tags will be removed at once. Hence I think an hardcoded label will be best suitable for this.

Still if you think the hardcoded label doesn't help, please elaborate with an example

shrilakshmishastry: Below are the cases where we need to consider adding a description to clear all tags:

1. multiple tags on a single page
2.  when Voice over user passes through this CTA, they just hear as Clear all . I have replicated what they feel here  https://codepen.io/shrilakshmishastry/pen/PoMWzwv

If you still feel like hardcoded value is enough I will add it

ad1992: Do you mean when you have multiple React Tags Component on a single page ? In that case, I agree the hard-coded aria label wouldn't be intuitive and we should provide an option to pass the aria label.

Even if we allow customizing the aria label via prop, we will need to come up with a better config for aria attributes instead of just adding prop for every element as the props list will keep growing. 

Let me think over this on how better we can pass aria props and get back to you. If you have any suggestions do share.

shrilakshmishastry: one way we can adopt is the below structure 
```
accessibilityProps?:{
labelOfclearAllCta?: string;
TagsLabelId?:string;
inputLabelId?:string;
}
```

shrilakshmishastry: let me know what you think.

ad1992: let's keep it short and simple. We can tweak before release if needed but I would like to unblock you for now

```
ariaAttrs?: {
  clearAllLabel?: string
}
```

shrilakshmishastry: lgtm. 
can we change clearAllLabel to clearAllButtonLabel to tell the label is for the button?

ad1992: I don't think its needed as its very intuitive and clear its for the the clear all button since there isn't any other element with the same functionality/label. So lets keep the prop short and simple

shrilakshmishastry: ok I will update the prop name and push the changes 👍

shrilakshmishastry: added above mentioned prop https://github.com/react-tags/react-tags/blob/b8590efe6ecfa94e8ea6d4b59b8666b0596a5b69/src/components/ReactTags.tsx#L44

src/components/ClearAllTags.tsx
```diff
@@ -4,10 +4,11 @@ const ClearAllTags = (props: {
   classNames: {
     clearAll: string,
   },
+  accessibilityLabel?: string,
   onClick: () => void,
 }) => {
   return (
-    <button className={props.classNames.clearAll} onClick={props.onClick}>
+    <button aria-label={props.accessibilityLabel} className={props.classNames.clearAll} onClick={props.onClick}>
```
ad1992, line 11: let's rename the prop to simply `aria-label` here

ad1992 (changes requested):

src/components/ClearAllTags.tsx
shrilakshmishastry: keeping it in camelCase and renamed to `ariaLabel ` https://github.com/react-tags/react-tags/blob/76cb9fffab90225487cb4bd2c614f0161e6d450a/src/components/ClearAllTags.tsx#L7

src/components/ClearAllTags.tsx
```diff
@@ -4,10 +4,11 @@ const ClearAllTags = (props: {
   classNames: {
     clearAll: string,
   },
+  ariaLabel?: string,
```
ad1992, line 7: In `React` all ARIA attribute names are the same as in HTML hence it is better we follow the same convention here as the DOM instead of camel case

ad1992 (changes requested):
Also lets update the readme for the new prop as well

shrilakshmishastry:
> doc: add description for ariaAttrs in README

Added 
https://github.com/react-tags/react-tags/blob/8471fa6e74bc50938dc091d1cf1f484eaa17a200/README.md?plain=1#L210

src/components/ClearAllTags.tsx
shrilakshmishastry: done https://github.com/react-tags/react-tags/blob/8471fa6e74bc50938dc091d1cf1f484eaa17a200/src/components/ClearAllTags.tsx#L7

README.md
```diff
@@ -700,3 +701,8 @@ This callback is if present is triggered when "clear all" button is clicked. You
 ### maxTags
 This prop specifies the maximum count of tags to be added. Incase the tags exceed, error will show up to convey the maximum tag limit has reached.
 
+### ariaAttrs
+
+An object used to pass custom accessibility attributes (ARIA properties) to enhance screen reader support and improve accessibility. For example, clearAllLabel defines the accessible label for a "clear all tags" button, providing context for assistive technologies.
```
ad1992, line 706:
```suggestion
An object containing custom ARIA attributes to enhance screen reader support and improve accessibility. 
It accepts the below attributes
```js
{
  clearAllLabel?: string // Optional accessible label for a "clear all tags" button improving accessibility by providing a clear, descriptive action label.
}
```

README.md
```diff
@@ -207,6 +207,7 @@ Option | Type | Default | Description
 |[`clearAll`](#clearAll) | `boolean` | `false` | Implies whether 'clear all' button should be shown.
 |[`onClearAll`](#onClearAll) | `Function` |  | This callback if present is triggered when clear all button is clicked.
 | [`maxTags`](#maxTags) | `number` | | The maximum count of tags to be added
+| [`ariaAttrs`](#ariaAttrs) | Object |`{clearAllLabel: "clear all tags"}` | An object used to pass custom accessibility attributes
```
ad1992, line 210:
```suggestion
| [`ariaAttrs`](#ariaAttrs) | Object |`{clearAllLabel: "clear all tags"}` | An object containing custom ARIA attributes to enhance screen reader support and improve accessibility
```

src/index.tsx
```diff
@@ -197,6 +197,13 @@ export interface ReactTagsWrapperProps {
    * Whether to display a button to clear all the tags.
    */
   clearAll?: boolean;
+  /**
+   * object to hold accessibility attributes for below list
+   *  - label for clear all button
+   */
+  ariaAttrs?: {
+    clearAllLabel?: string
+  };
```
ad1992, line 206:
```suggestion
  /**
   * An object containing custom aria attributes to improve acceessibility
   */
  ariaAttrs?: {
  // label for clear all button
    clearAllLabel?: string
  };
```

ad1992 (commented):
@shrilakshmishastry looks good, I have made some tweaks. One last thing, lets add a test as well :)

shrilakshmishastry:
> @shrilakshmishastry looks good, I have made some tweaks. One last thing, lets add a test as well :)



> @shrilakshmishastry looks good, I have made some tweaks. One last thing, lets add a test as well :)

Added test case
https://github.com/react-tags/react-tags/blob/f6aaa4d2cac93b8a57f7a2ed270bb645e5b81b4e/__tests__/ReactTags.test.tsx#L1307

__tests__/ReactTags.test.tsx
```diff
@@ -1304,20 +1304,65 @@ describe('Test ReactTags', () => {
...
+          aria-label="clear all tags"
           class="ReactTags__clearAll"
         >
           Clear all
         </button>
       `);
+
+    })
+
+      it('should have  value for aria-label when label is provided', () => {
```
ad1992, line 1326:
```suggestion
      it('should have value for aria-label when label is provided', () => {
```

ad1992 (approved):

ad1992:
Awesome, merging! Thanks @shrilakshmishastry

PR #1002, opened 2024-10-02
feat: accept labelledById prop to provide label for input tags

shrilakshmishastry:
## ADDED
 `labelledById`: a required prop. This is the ID of the label of the input field and the tag container

## FIX

Issue: For assistive technology users 
  1. when they input value to a  tag name, they will not have any context on what they are inputting for
  2. When they edit pre-added tags they will not know what they are editing and the context of tags.

Solution : 
1. accept a mandatory prop `labelledById` an id to label for both tag and input. 

Screenshot VO when editing tag

Screenshot 2024-10-02 at 11 06 48 AM

accessibility tree when editing tag
|then|now|
|--|--|
|Screenshot 2024-10-02 at 11 46 39 AM|Screenshot 2024-10-02 at 11 07 04 AM (2)|
| Screenshot 2024-10-02 at 11 46 27 AM| image|

screenshot of VO on input 
Screenshot 2024-10-02 at 11 05 57 AM (2)

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -12,6 +12,10 @@ import {
 
 export interface ReactTagsWrapperProps {
   /**
+   * Id of the label of the input field and the tags container.
+   */
+  labelledById: string;
+  /**
    * Placeholder text for the input field.
    */
   placeholder?: string;
@@ -248,10 +252,12 @@ const ReactTagsWrapper = (props: ReactTagsWrapperProps) => {
     inputValue,
     maxTags,
     renderSuggestion,
+    labelledById,
   } = props;
 
   return (
     <ReactTags
+      labelledById={labelledById}
       placeholder={placeholder}
       labelField={labelField}
       suggestions={suggestions}
--- a/example/main.tsx
+++ b/example/main.tsx
@@ -67,9 +67,10 @@ const App = () => {
   return (
     <div className="app">
       <GitHubCorner />
-      <h1> React Tags Example </h1>
+      <h1 id="react-tags-example"> React Tags Example </h1>
       <div>
         <ReactTags
+          labelledById="react-tags-example"
           tags={tags}
           suggestions={suggestions}
           separators={[SEPARATORS.ENTER, SEPARATORS.COMMA]}
--- a/src/components/ReactTags.tsx
+++ b/src/components/ReactTags.tsx
@@ -41,6 +41,7 @@ type ReactTagsProps = ReactTagsWrapperProps & {
   inputProps: { [key: string]: string };
   editable: boolean;
   clearAll: boolean;
+  labelledById: string;
 };
 
 const ReactTags = (props: ReactTagsProps) => {
@@ -70,6 +71,7 @@ const ReactTags = (props: ReactTagsProps) => {
     maxLength,
     inputValue,
     clearAll,
+    labelledById,
   } = props;
 
   const [suggestions, setSuggestions] = useState(props.suggestions);
@@ -469,6 +471,7 @@ const ReactTags = (props: ReactTagsProps) => {
                 }}
                 onFocus={handleFocus}
                 value={query}
+                aria-label={`Editing tag ${query} of ${labelledById}`}
                 onChange={handleChange}
                 onKeyDown={handleKeyDown}
                 onBlur={handleBlur}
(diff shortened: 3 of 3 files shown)


src/index.tsx
```diff
@@ -248,10 +252,12 @@ const ReactTagsWrapper = (props: ReactTagsWrapperProps) => {
     inputValue,
     maxTags,
     renderSuggestion,
+    labelledById,
```
ad1992, line 255: @shrilakshmishastry why this needs to be a user prop ? We can hardcode it instead ?

Ww can append a random id to it so that if there are multiple tags, every tag component has a unique id. We can generate the id using [crypto.randomUUID](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) and shorten its length

```
const id = crypto.randomUUID().slice(0, 8)
```

ad1992 (changes requested):

ad1992:
Also @shrilakshmishastry please merge the branch with master so the conflicts are gone :)

src/index.tsx
shrilakshmishastry: Please explain a bit more about hardcoding part.

Context of labelledById here : 
This is the title or label given to describe the context in which tag is being used. In the example set up we have, it's `react tags example`. 

To describe how a screen reader user feels when he goes through the tags without the label being add is he reads heading `react tags example` then read tags lets say India is first one, it will be read as `India`. India can be taken as description also .

shrilakshmishastry: After reading the second sentence in your comment, I felt that you're considering labeling the tag when it's not been edited. 

I am adding the label to tag when it is being edited 😄

ad1992: I meant hardcoding the label as `React Tags`. lets do this, have a default value `React Tags`, it shouldn't be a required prop, we should have always have a default/fallback value.
Additonally I realized the prev json structure we decided isn't the best so lets improve that and instead of `ariaAttrs` based on the component.

so for clear all button it would be 

```
clearAllBtnAriaAttrs?: {
    aria-label?: string
}
```

And for input it will be `inputAria Attrs`

```
inputAriaAttrs?: {
    aria-labelledby?:string
 }
```

This structure looks more clean and has separation for components. Wdyt?

__tests__/ReactTags.test.tsx
```diff
@@ -35,6 +35,9 @@ beforeAll(() => {
       { className: '', id: 'Peach', text: 'Peach' },
     ],
     handleDrag: handleDragStub,
+    ariaAttrs: {
+            labelledById:''
+          }
```
ad1992, line 40: looks like not formatted

src/index.tsx
ad1992: > I agree with the prop-type structure. But I am not ok with the label being React tags. Let's say I have tags of countries, how come React tags make sense here?

That would be the default label, we can simplify it to `tags` or some other name if you have better suggestions :)
We need a default generic label that would be used in case the users don't pass the label explicitly

__tests__/ReactTags.test.tsx
shrilakshmishastry: formatted now https://github.com/react-tags/react-tags/blob/7fbd38bde1a6b3be2e592b6516e16a330fa05f9e/__tests__/ReactTags.test.tsx#L38

src/index.tsx
shrilakshmishastry: re-structured the props now to accept accessibility attributes based on component

https://github.com/react-tags/react-tags/blob/7fbd38bde1a6b3be2e592b6516e16a330fa05f9e/src/index.tsx#L205

https://github.com/react-tags/react-tags/blob/7fbd38bde1a6b3be2e592b6516e16a330fa05f9e/src/index.tsx#L214

Made the default label as tags https://github.com/react-tags/react-tags/blob/7fbd38bde1a6b3be2e592b6516e16a330fa05f9e/src/components/ReactTags.tsx#L530

README.md
```diff
@@ -701,13 +702,29 @@ This callback is if present is triggered when "clear all" button is clicked. You
...
+```
+
+
+### inputAriaAttrs
+
+An object containing accessibility-related attributes for the `<input>` field and tags container, designed to improve screen reader compatibility and enhance user experience for individuals using assistive technologies.
+
+`aria-labelledby` (optional): The ID of the label associated with both the input field and the tags container. This attribute links the label and input, providing context to screen readers on what the input represents.
 
+`titleLabel` (optional): A label string that alerts screen reader users when they are editing a tag associated with a specific heading. This helps users understand the context of the tag they’re currently editing.
```
ad1992, line 722: This should be `aria-label`, lets use the exact same HTML attribute which needs to be passed to the react components

ad1992 (commented):
One last change @shrilakshmishastry and we are good to go

PR #1020, opened 2026-06-10
fix: restore suggestions after Escape

### LINKED ISSUE #1018: Suggestion List Does Not Reappear After Pressing Escape
# Description

When pressing the `Escape` key in the tag input field, the suggestion list disappears as expected. However, it does not reappear when the window or tag input field regains focus.

I believe pressing `Escape` should behave like clicking outside the tag input field: the suggestion list is hidden but should reappear when the input field is clicked again.

# Steps to Reproduce

1. Open the [Sandbox for this project](https://codesandbox.io/p/sandbox/react-tags-l65545).
2. Type `Ind` in the tag input field — the suggestion list appears with options like "India", "Indonesia", etc.
3. Click outside the tag input field — the suggestion list disappears.
4. Click inside the tag input field again — the suggestion list reappears with the correct suggestions.
5. Press the `Escape` key — the suggestion list disappears (as expected).
6. Click outside the tag input field, then click back inside — the suggestion list does not reappear ❌
7. Type or delete a character in the input field — the suggestion list reappears ✅

A [recording](https://github.com/user-attachments/assets/19892509-22be-4ea5-a0c2-5bf625ceefb4) demonstrates the above steps.

# Expected behavior

After pressing the Escape key, clicking back into the tag input field should immediately show the suggestion list again, consistent with steps 2–4.

# Tested on

Tested on v6.10.6(2025-02-18)

xub5210-droid:
Fixes #1018

Recomputes filtered suggestions when the input regains focus and adds a regression test covering Escape, blur, and refocus.

Testing: regression test added; git diff --check. Full Jest suite was not run because dependency installation did not complete locally.


xub5210-droid:
The failing Codacy coverage check appears to be unrelated to this PR.

The workflow successfully completes the first `actions/checkout@master` step, but then fails during the duplicate `actions/checkout@v2` step with an HTTP 400 response. Dependency installation and tests are skipped as a result.

The other checks, including the main workflow and Codacy static analysis, passed. Could a maintainer please rerun the failed workflow or update/remove the duplicate `actions/checkout@v2` step?

